PR #1, opened 2024-08-16
RAG Web Browser implementation

jirispilka:
Description
- **Inputs**: I aim to keep the configuration minimal yet usable (limit to EN, US for now).
- **Functionality**: Only extract titles and URLs from Google search results (adapted from Google extractors).
- **Cheerio**: Used for Google Search to improve speed.
- **Playwright**: Used for scraping (might be slow, but JavaScript rendering is a requirement). I avoided complicating this with adaptive crawling or crawler selection. We can keep it for later.
**Output**: Configurable options include **text** (default), **HTML** (optional), and **Markdown** (optional).

see [slack discussion](https://apify.slack.com/archives/C04URRT3934/p1722953954846909) for a reference


jirispilka:
I need to change this actor to incorporate standby mode, as I initially misunderstood the standby functionality. Originally, I thought that standby mode would automatically wrap the existing actor into an "API server", exposing the running container. Therefore, I expected that sending a POST request to container_url with the payload would work.

jancurn:
Good work @jirispilka, here is feedback from my side:

> * **Name**: I asked Ales W. for a good name (to avoid overlaps with the existing actors)

How about **RAG Web browser (rag-web-browser)** ?

> * **Inputs**: I aim to keep the configuration minimal yet usable (limit to EN, US for now).

Absolutely, I was thinking about INPUT like this:

- I’d use `query` instead of `queries`, just to keep it simple and nice in code example. In Standby API, you can call this endpoint multiple times for multiple queries anyway, which better balances the performance.
- `maxPages` - Maximum resulting number of web pages to load.
- FUTURE: `searchType` (select: search|images|..) -  Would be great to expand to Google Image search too, now that "[Traditional or "naive" Retrieval-Augmented Generation (RAG) is dead"](https://apify.slack.com/archives/C04URRT3934/p1724349413531479?thread_ts=1724091940.070259&cid=C04URRT3934) :) 
- `formats` (multi-option: `text|markdown|html|image|pdf|docx|file`) - Select text formats to be included in the results. 
- `pageTimeoutSecs` - How many seconds wait for loading target web page. If reached, it will be skipped in the results. This is to ensure we’re within ~40 seconds timeout for function calling in GPTs and elsewhere. Default: 30 seconds.

The Standby API should just replicate input 1:1, and expose GET API endpoint at e.g. `/search?query=test&formats=html,markdown&pageTimeoutSecs=5&...`

> * **Functionality**: Only extract titles and URLs from Google search results (adapted from Google extractors).

Sounds great

> * **Cheerio**: Used for Google Search to improve speed.

Yeah, makes sense.

> * **Playwright**: Used for scraping (might be slow, but JavaScript rendering is a requirement). I avoided complicating this with adaptive crawling or crawler selection. We can keep it for later.

Yeah absolutely, also if we use just one browser (Firefox?), it will keep the Docker image small and thus fast to start.

>   **Output**: Configurable options include **text** (default), **HTML** (optional), and **Markdown** (optional).

Sounds good, the results could look like this (either stored to output dataset, or returned directly from Standby API):

```js
{
[ 
    {
      text: “...chunk of text from target page ...”,
	markdown: "...", // if enabled 
      metadata: {
        url: “http://www.example.com/dir/page.htm“,
        // createdAt: ...ISODate, // date of creation, if available
        // author: "Bob Dylan", // Author of page, if available
      },
    },
    ...
  ],
}
```


> 
> TODOs:
> 
> * add standby mode
> * extract PDF, Docx content?

Yeah, definitely, we just need to figure how are we going to return file content in Standby API, in Base64 or kv-store?

> see [slack discussion](https://apify.slack.com/archives/C04URRT3934/p1722953954846909) for a reference

jirispilka:
@jancurn Thanks for the feedback!
> 
> > * **Name**: I asked Ales W. for a good name (to avoid overlaps with the existing actors)
> 
> How about **RAG Web browser (rag-web-browser)** ?
> 
Ales has suggested: `Google Search Data Extractor`, which I like. I think it would be good to also include the "web browser"

> * I’d use `query` instead of `queries`, just to keep it simple and nice in code example. In Standby API, you can call this endpoint multiple times for multiple queries anyway, which better balances the performance.

`queries` I kept the same as for the google-search-actor, but I will change it to `query` (sounds better)

> * `maxPages` - Maximum resulting number of web pages to load.

I don't want to confuse users whether this refers to number of Google search pages to crawl or number of Google search results to return.  I think `maxResults` is more clear in that sense. @jancurn what do you think?

> * `formats` (multi-option: `text|markdown|html|image|pdf|docx|file`) - Select text formats to be included in the results.
> * `pageTimeoutSecs` - How many seconds wait for loading target web page. If reached, it will be skipped in the results. This is to ensure we’re within ~40 seconds timeout for function calling in GPTs and elsewhere. Default: 30 seconds.
> 
> The Standby API should just replicate input 1:1, and expose GET API endpoint at e.g. `/search?query=test&formats=html,markdown&pageTimeoutSecs=5&...`

Thanks, will do

> > * **Playwright**: Used for scraping (might be slow, but JavaScript rendering is a requirement). I avoided complicating this with adaptive crawling or crawler selection. We can keep it for later.
> 
> Yeah absolutely, also if we use just one browser (Firefox?), it will keep the Docker image small and thus fast to start.

yes, firefox

> > **Output**: Configurable options include **text** (default), **HTML** (optional), and **Markdown** (optional).
> 
> Sounds good, the results could look like this (either stored to output dataset, or returned directly from Standby API):
> 
> ```js
> {
> [ 
>     {
>       text: “...chunk of text from target page ...”,
> 	markdown: "...", // if enabled 
>       metadata: {
>         url: “http://www.example.com/dir/page.htm“,
>         // createdAt: ...ISODate, // date of creation, if available
>         // author: "Bob Dylan", // Author of page, if available
>       },
>     },
>     ...
>   ],
> }
> ```
ok, I'm having flat structure now, i.e.
```
{ 
  "title": "",
  "url" : "".
  "text": .......
}
```
Good point with the metadata. If we want to store additional info, it would be better to create the metadata fiedl right away.

jancurn:
- "Google Search Data Extractor" is way too close to "Google Search Results Scraper" and all the other Google Search scrapers we already have in Store (see https://apify.com/store/categories?search=google+search), plus it feels like it just extracts data from Google Search, which is not true - the data is in web pages! And it doesn't communicate the main application of this Actor - for being web browser for RAG. Let's go with **RAG Web browser (`apify/rag-web-browser`)** 
- re `maxResults` - fair enough, the results can be pages, but also files or images in the future
- "ok, I'm having flat structure now, i.e." - we need to return results from more than one page, so it needs to be an array. Good point about adding `title` too.

jancurn:
For inspiration of results, see https://tavily.com/#api

src/main.ts
```diff
@@ -0,0 +1,84 @@
...
+    } catch (e) {
+        const error = e as Error;
+        const errorMessage = { errorMessage: error.message };
+        const statusCode = error instanceof UserInputError ? 400 : 500;
+        res.writeHead(statusCode, { 'Content-Type': 'application/json' });
+        res.end(JSON.stringify(errorMessage));
+    }
+});
+
+if ((Actor.isAtHome() && Actor.getEnv().metaOrigin === 'STANDBY') || RUN_STANDBY_MODE_AT_LOCAL) {
```
metalwarrior665, line 54: You don't need (Actor.isAtHome() and RUN_STANDBY_MODE_AT_LOCAL, you can pass it env var locally when running

```
APIFY_META_ORIGIN=STANDBY apify run
```

src/extractors.ts
```diff
@@ -0,0 +1,43 @@
+import { Readability, isProbablyReaderable } from '@mozilla/readability';
```
metalwarrior665, line 1: I would put all the extractor/util code that you took from other Actors into a folder so readers can easily see which code we might later want to put somewhere to import from (will handle the `google-extractors` library so it auto publishes)

src/crawlers.ts
```diff
@@ -0,0 +1,138 @@
...
+// Proxy configuration for a PlaywrightCrawler
+// How to handle the case when all the requests are not finished?
+
+export async function createAndStartSearchCrawler(
+    crawlerOptions: CrawlerOptions = {} as CrawlerOptions,
+    scraperSettings: ScraperSettings = {} as ScraperSettings,
+    startCrawler: boolean = true,
+) {
+    const proxyConfig = await Actor.createProxyConfiguration(crawlerOptions.proxyConfigurationOptions);
+
```
metalwarrior665, line 51: Is there a reason you are not using the [memory storage as in the SuperScraper](https://github.com/apify/super-scraper/blob/master/src/crawlers.ts#L19)? It is a tiny bit faster, you don't need to use a named queue (since the queues don't collide) that you would need to delete. We use memory storage because after migration you loose the connection anyway so users need to resend the request, there is no need to persist anything.

src/crawlers.ts
```diff
@@ -0,0 +1,138 @@
...
+/**
+ * Adds a search request to the Google search crawler.
+ * Create a response for the request and set the desired number of results (maxResults).
+ */
+export const addSearchRequest = async (
+    request: RequestOptions<UserData>, response: ServerResponse | null,
+    crawlerOptions: CrawlerOptions, maxResults: number,
+) => {
+    const key = CrawlerType.CHEERIO_GOOGLE_SEARCH_CRAWLER;
```
metalwarrior665, line 118: I would add a comment that currently this getting of crawler from map by the key is not doing anything since we pre-create the two crawlers anyway but later it can be used e.g. if user wants to use different proxy configs

metalwarrior665 (commented):
Looks great!

src/main.ts
jirispilka: oh, ok. Thanks!

src/crawlers.ts
jirispilka: Good point! I had tried separating the queues earlier, but for some reason, they were colliding, and I couldn't figure out why.

Here's the relevant code:
```
    const client = new MemoryStorage();
    const queue = await RequestQueue.open(undefined, { storageClient: client });

    const crawler = new CheerioCrawler({
        ...(cheerioCrawlerOptions as CheerioCrawlerOptions),
        requestQueue: queue,
```
and
```
    const crawler = new PlaywrightCrawler({
        ...(crawlerOptions as PlaywrightCrawlerOptions),
        requestQueue: await RequestQueue.open(),
```
The issue is that the second queue ends up with the same ID and the same client, causing the queues to collide.

@metalwarrior665 Can you please help with that (there must be some fundaments I'm missing). I added it into this branch [feat/replace-queue](https://github.com/apify/rag-web-browser/blob/feat/replace-queue/src/crawlers.ts)

src/extractors.ts
jirispilka: I move the relevant code to the `google-search` and `website-content-crawler` directories to improve the readability, please check in the branch [feat/create-extractor-dir](https://github.com/apify/rag-web-browser/tree/feat/create-extractor-dir/src)

Although, I'm not sure how it can be imported and auto-published in the future.

src/crawlers.ts
metalwarrior665: I was debugging this and it seems the only way to make this work is to make both queues with memory storage AND gives them different names.

metalwarrior665: Also add `
const client1 = new MemoryStorage({ persistStorage: false});`

src/extractors.ts
metalwarrior665: For Google Search we have the package, I just didn't automate it yet https://www.npmjs.com/package/@apify/google-extractors

We could even update this Actor's repo and rebuild it automatically but it doesn't break as often so we will see how many users you will have.

jirispilka: ok, thanks for the explanation

src/crawlers.ts
jirispilka: Thanks, it is working fine. Done

src/extractors.ts
jancurn: FYI, the plan was to make this repo open source, hence it shouldn't depend on external private packages. For this use case (simply extracting titles + URLs from organic google search results), we thought simple CSS selector will do the job, without a need for the full-featured google extractor.

jirispilka: @jancurn that's correct. Currently, we're not using the google-extractors library; instead, we're use css selectors to extract links from search results

metalwarrior665: Yeah, that makes sense. The extractors broke like once or twice per year so you can manually update it based on our changes.

jancurn: Cool, the name of the directories confused me a bit :)

jirispilka:
@metalwarrior665 @jancurn I would like to ask you again for review. 

Performance-wise, there is still plenty of room for an improvement. I did ad-hoc testing, with the following summary:

Here is a typical latency breakdown for the RAG Web Browser.
Please note, the these results are only indicative and may vary based on the search term and the target websites.

The numbers below are based on the following search terms: "apify", "Donald Trump", "boston". Results were averaged for the three queries.

| Memory (GB) | Max Results | Latency (s) |
|-------------|-------------|-------------|
| 2           | 1           | 36          |
| 2           | 5           | 88          |
| 4           | 1           | 22          |
| 4           | 5           | 46          |

You can find a breakdown in the [performance_measures.md](https://github.com/apify/rag-web-browser/blob/feat/actor-impl/data/performance_measures.md)

```text
Average:

request-received: 0 s
before-cheerio-queue-add: 129
cheerio-request-handler-start: 2499
before-playwright-queue-add: 87
playwright-request-start: 15703
playwright-wait-dynamic-content: 7029
playwright-remove-cookie: 1073
playwright-parse-with-cheerio: 5564
playwright-process-html: 3829
playwright-before-response-send: 236 
Time taken for each request: [ 49762, 16004, 42676 ]
```

There are some low-hanging fruits like: `playwright-wait-dynamic-content`, `playwright-remove-cookie`.
However, I'm uncertain about the potential improvements for other things like `playwright-request-start`.

jirispilka:
> Great job. 2 GB is definitely suboptimal for ideal speed. I think you could even go higher than 4 GB for default, depending if cost or latency is a bigger priority.

I updated the README in this spirit
```
Based on your requirements, if low latency is a priority, consider running the Actor with 4GB or more of memory.
However, if you're looking for a cost-effective solution, you can run the Actor with 2GB of memory.
``` 
> How many pages you think it will scrape usually? If 5 is normal, you could theoretically hyper-optimize by spawning those in separate standby Actors

I'm not sure about how many pages people might use. My guess was somewhere around 1 to 5. I'm starting playwright with `desiredConcurrency = 3`.

jirispilka:
Let me merge this PR now for the following reasons:
- The Actor has already been published, so it’s better to have the code in the master branch for anyone visiting the repository.
- This PR has become quite lengthy and complex.

PR #26, opened 2024-10-31
feat: Jan copy (improved README, simplified configuration, handle validation)

jancurn:


jirispilka:
@jancurn Thanks a lot! I like the improved structure, it’s definitely better now. Usage -> Normal Mode and Usage -> Standby Mode, cool.

I also like the feature list with emojis as it adds a nice touch.

I've included an Output example at the top so users can quickly see what they’ll get.
I also included a link to the OpenAPI schema for Standby Mode rather than pasting the entire schema, as it would take up too much space in the README and could be distracting. I don’t have a strong opinion on this, as I assume people use the right-side navigation for quickly browsing the README. 

Let me know, I can change that. 

Otherwise we can merge and publish.

jirispilka:
@jancurn are you planning to do more work on this PR?

PR #27, opened 2024-11-06
feat: handle url in a query

jirispilka:
I’ve added functionality to handle both full URLs and simple domains, such as "apify.com," in the query.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -5,20 +5,6 @@ import { v4 as uuidv4 } from 'uuid';
 import defaults from './defaults.json' with { type: 'json' };
 import { OrganicResult, TimeMeasure, UserData } from './types.js';
 
-const PROTOCOL_REGEX = /^(https?|ftp):\/\//;
-const DOMAIN_REGEX = /^([a-zA-Z0-9-]+\.)+[a-zA-Z]{2,6}(\/[a-zA-Z0-9._~:/?#[\]@!$&'()*+,;=%-]*)?$/;
-
-const URL_REGEX = new RegExp(
-    '^'
-    + '(https?:\\/\\/|ftp:\\/\\/)?' // Optional protocol (http://, https://, ftp://)
-    + '(([a-zA-Z0-9-]+\\.)+[a-zA-Z]{2,6}' // Domain name with TLD (e.g., example.com)
-    + '|'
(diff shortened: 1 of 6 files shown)


src/utils.ts
```diff
@@ -5,6 +5,20 @@ import { v4 as uuidv4 } from 'uuid';
 import defaults from './defaults.json' with { type: 'json' };
 import { OrganicResult, TimeMeasure, UserData } from './types.js';
 
+const PROTOCOL_REGEX = /^(https?|ftp):\/\//;
+const DOMAIN_REGEX = /^([a-zA-Z0-9-]+\.)+[a-zA-Z]{2,6}(\/[a-zA-Z0-9._~:/?#[\]@!$&'()*+,;=%-]*)?$/;
+
+const URL_REGEX = new RegExp(
```
jancurn, line 11: Do we really need this? Why not use `URL.parse()`? BTW we should only consider it a URL if it starts with `http://` or `https://` (and not others like `ftp://`), e.g. consider "test.io" a search phrase, not URL

src/utils.ts
```diff
@@ -91,3 +111,25 @@ export function transformTimeMeasuresToRelative(timeMeasures: TimeMeasure[]): Ti
...
+ * Convert a string to a proper URL.
+ */
+export function toProperUrl(url: string) {
+    // If it starts with a valid protocol, return as is
+    if (PROTOCOL_REGEX.test(url)) {
+        return url;
+    }
+
+    // If it matches a domain pattern, prepend http or https
+    if (DOMAIN_REGEX.test(url)) {
```
jancurn, line 129: Same here, let's use JS `URL` class rather than our code

jancurn (changes requested):

src/utils.ts
```diff
@@ -112,24 +98,11 @@ export function transformTimeMeasuresToRelative(timeMeasures: TimeMeasure[]): Ti
...
-export function toProperUrl(url: string) {
-    // If it starts with a valid protocol, return as is
-    if (PROTOCOL_REGEX.test(url)) {
-        return url;
-    }
-
-    // If it matches a domain pattern, prepend http or https
-    if (DOMAIN_REGEX.test(url)) {
-        return `https://${url}`;
+export function safeUrl(input: string): URL | null {
```
jancurn, line 101: I'd call this function something like `interpretAsUrl`, return null if it's not http/https, and always return a string.

jancurn (approved):

PR #32, opened 2024-11-18
Jan/changes

jancurn:


jirispilka:
@jancurn Thanks a lot for changes. I did a couple of modification and tested in the console. Please check and then we can merge it.

PR #45, opened 2025-02-04
fix: Handle double encoding of URL

jirispilka:


src/utils.ts
```diff
@@ -110,12 +110,26 @@ export function transformTimeMeasuresToRelative(timeMeasures: TimeMeasure[]): Ti
...
+            return null;
+        }
+    }
+
+    let candidate = input;
+    for (let i = 0; i < 3; i++) {
+        const result = tryValid(candidate);
+        if (result) return result;
+        try {
+            candidate = decodeURIComponent(candidate);
```
jancurn, line 129: Comment would be nice

jancurn (approved):

MQ37 (approved):
LGTM :+1:

src/utils.ts
jirispilka: Thanks, I was a bit lazy 😊 . Fixed

PR #51, opened 2025-02-28
Fix: Standby mode spawns extra crawlers

### LINKED ISSUE #50: Standby mode spawns extra crawlers
While working on adding cheerio content crawler (#48) I have noticed that handling crawlers isn't working properly. When the Actor runs in a standby mode it will correctly create both of the crawlers but every time a search request is run with different query parameters new Playwright crawler is created. I suspect that this is caused by this method in `crawlers.ts`:

```ts
export function getPlaywrightCrawlerKey(
    playwrightCrawlerOptions: PlaywrightCrawlerOptions,
    playwrightScraperSettings: PlaywrightScraperSettings,
) {
    return JSON.stringify(playwrightCrawlerOptions) + JSON.stringify(playwrightScraperSettings);
}
```

It uses `playwrightScraperSettings` as part of the key even though it could change between requests and has no effect on the crawler instance itself.

Here is a log from simple run with one request: https://gist.github.com/matyascimbulka/0e74bf8597e3b673268cf6564f94fb69

matyascimbulka:
Fixes #50.

Refactors the PlaywrightScraperSettings into UserData. This reduces the amount of create crawlers. New crawlers are only created when changing settings related directly to the crawler (eg. `maxRequestRetries`).

I have also removed pre-creation of crawlers in standby mode as the requested crawlers could have different settings and the overhead of creating crawlers is small.

cc: @metalwarrior665 @jirispilka

### DIFF
--- a/src/const.ts
+++ b/src/const.ts
@@ -29,7 +29,7 @@ export const defaults = {
     minConcurrency: inputSchema.properties.minConcurrency.default,
     outputFormats: inputSchema.properties.outputFormats.default,
     proxyConfiguration: inputSchema.properties.proxyConfiguration.default,
-    query: undefined, // No defult value in input_schema.json
+    query: undefined, // No default value in input_schema.json
     readableTextCharThreshold: 100, // Not in input_schema.json
     removeCookieWarnings: inputSchema.properties.removeCookieWarnings.default,
     removeElementsCssSelector: inputSchema.properties.removeElementsCssSelector.default,
--- a/src/const.ts
+++ b/src/const.ts
@@ -1,3 +1,5 @@
+import inputSchema from '../.actor/input_schema.json' assert { type: 'json' };
+
 export enum ContentCrawlerStatus {
     PENDING = 'pending',
     HANDLED = 'handled',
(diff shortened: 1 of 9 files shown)


src/crawlers.ts
```diff
@@ -97,7 +88,12 @@ async function createAndStartSearchCrawler(
             let rank = 1;
             for (const result of results) {
                 result.rank = rank++;
-                const r = createRequest(result, responseId, request.userData.timeMeasures!);
+                const r = createRequest(
+                    result,
+                    responseId,
+                    request.userData.playwrightScraperSettings!,
+                    request.userData.timeMeasures!,
+                );
```
metalwarrior665, line 96: I would refactor this more in future PR, at least it should have named parameters, this is already hard to order. But let's not do it in this PR

src/playwright-req-handler.ts
```diff
@@ -59,11 +59,14 @@ function isValidContentType(contentType: string | undefined) {
...
-export async function requestHandlerPlaywright(
-    context: PlaywrightCrawlingContext<UserData>,
-    settings: PlaywrightScraperSettings,
-) {
+export async function requestHandlerPlaywright(context: PlaywrightCrawlingContext<UserData>) {
     const { request, contentType, page, response, closeCookieModals } = context;
+    const { playwrightScraperSettings: settings, responseId } = request.userData;
+
+    if (!settings) {
+        log.error('Missing playwrightScraperSettings in request.userData');
```
metalwarrior665, line 67: If you use typed UserData then TS should ensure you pass it in and you can remove this error. Just make sure whenever you addRequest, you also use the UserData type. If you would need to make separate UserData for each Crawler then I would do it in this PR, less chance we will mess something up and need this error

src/main.ts
```diff
@@ -75,9 +74,7 @@ if (standbyMode) {
     const host = Actor.isAtHome() ? process.env.ACTOR_STANDBY_URL : 'http://localhost';
     const port = Actor.isAtHome() ? process.env.ACTOR_STANDBY_PORT : 3000;
     app.listen(port, async () => {
-        // Pre-create default crawlers
         log.info(`The Actor web server is listening for user requests at ${host}.`);
-        await createAndStartCrawlers(cheerioCrawlerOptions, playwrightCrawlerOptions, playwrightScraperSettings);
```
metalwarrior665, line 80: Please keep this in. This Actor is latency-critical. Even if this shaves just few millis, there is no reason to not do it. Most users will run with the default Crawler settings so they should use the preloaded.

metalwarrior665 (changes requested):
Thanks, looks good, just small things

src/main.ts
matyascimbulka: Then we need to sync the default query parameters and default input values. Because right now the first query will spawn new crawler because it has different settings. Look at this [gist](https://gist.github.com/matyascimbulka/0e74bf8597e3b673268cf6564f94fb69) (lines 17 - 21) where the 3rd crawler is spawn when calling the actor with `/search?query=crawlee`.

jirispilka: This is a good catch. I can see the values are indeed different in const.ts and input_schema.json

It should read, in `const.ts`
```
    initialConcurrency: 3,
    maxConcurrency: 50,
    minConcurrency: 1,
```
In `input_schema.json`
```
        "initialConcurrency": {
            "title": "Initial browsing concurrency",
            "type": "integer",
            "description": "The initial number of web browsers running in parallel. The system automatically scales the number based on the CPU and memory usage, in the range specified by `minConcurrency` and `maxConcurrency`. If the initial value is `0`, the Actor picks the number automatically based on the available memory.",
            "minimum": 0,
            "maximum": 50,
            "default": 3,
            "editor": "hidden"
        },
        "minConcurrency": {
            "title": "Minimum browsing concurrency",
            "type": "integer",
            "description": "The minimum number of web browsers running in parallel.",
            "minimum": 1,
            "maximum": 50,
            "default": 1,
            "editor": "hidden"
        },
        "maxConcurrency": {
            "title": "Maximum browsing concurrency",
            "type": "integer",
            "description": "The maximum number of web browsers running in parallel.",
            "minimum": 1,
            "maximum": 100,
            "default": 50,
            "editor": "hidden"
        },    
```        

Perhaps, a better way would be to remove it from `const.ts` altogether and read it from `input_schema.json` but is it a good idea?

src/utils.ts
```diff
@@ -72,6 +73,7 @@ export function createSearchRequest(
 export function createRequest(
     result: OrganicResult,
     responseId: string,
+    playwrightScraperSettings: PlaywrightScraperSettings,
```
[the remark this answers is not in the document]
matyascimbulka: I'm not sure how to fix this because I'm not getting this error. But I have tried removing `type` keyword from the import statement for the `types.ts` file (which I added earlier on).

matyascimbulka:
I have implemented a couple of changes:
1. Parsing of input_schema.json at startup so we have only one source of truth for default values.
2. Separated the UserData types for each crawler. It works as expected but it has introduced type error in `crawlers.ts` on line 138. Which I'm not quite sure how to deal with. Any help would be appreciated.

src/const.ts
```diff
@@ -12,29 +14,28 @@ export enum Routes {
...
+    keepAlive: true, // Not in input_schema.json
+    maxConcurrency: inputSchema.properties.maxConcurrency.default,
+    maxRequestRetries: inputSchema.properties.maxRequestRetries.default,
+    maxRequestRetriesMax: inputSchema.properties.maxRequestRetries.maximum,
+    maxResults: inputSchema.properties.maxResults.default,
+    maxResultsMax: inputSchema.properties.maxResults.maximum,
+    minConcurrency: inputSchema.properties.minConcurrency.default,
+    outputFormats: inputSchema.properties.outputFormats.default,
+    proxyConfiguration: inputSchema.properties.proxyConfiguration.default,
+    query: undefined, // No defult value in input_schema.json
```
jirispilka, line 32:
```suggestion
    query: undefined, // No default value in input_schema.json
```

src/utils.ts
```diff
@@ -59,33 +60,38 @@ export function createSearchRequest(
...
     };
 }
 
 /**
  * Create a request for Playwright crawler with the provided result, responseId and timeMeasures.
+ * @param query
  * @param result
  * @param responseId
  * @param timeMeasures
  */
```
jirispilka, line 73:
```suggestion
 * Create a request for Playwright crawler with the provided result, responseId and timeMeasures.
 */
```

jirispilka (approved):
Looks good to me, but I'm unsure about the type error. Perhaps Lukas will be able to help.

This reminds me, should we add github workflow for lint, test checks?

src/const.ts
matyascimbulka: Thanks for noticing, it's fixed.

src/utils.ts
matyascimbulka: Refactored

src/main.ts
metalwarrior665: We could read it but we need to make it non-fallible (we don't want to crash if someone moves it) and it would fallback to some defaults anyway. I would just sync it the values now and add a comment that we could read the schema.

metalwarrior665 (changes requested):
Not sure about the type error, you can work around with `requestHandler: requestHandlerPlaywright as unknown as PlaywrightRequestHandler,`

src/const.ts
```diff
@@ -1,3 +1,5 @@
+import inputSchema from '../.actor/input_schema.json' assert { type: 'json' };
```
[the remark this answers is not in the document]
matyascimbulka: When the file is moved/renamed `tsc` throws this error: `Cannot find module '../.actor/input_schema.json' or its corresponding type declarations.`. So I assume its safe to leave it as is.

Issue #48, opened 2025-02-18
Consider to add CheerioCrawler for the web pages


jirispilka:
Context: https://apify.slack.com/archives/C05683VTD6J/p1738881355331169

jirispilka:
CImplemented in #52

PR #52, opened 2025-03-04
feat: Add Cheerio content crawler

matyascimbulka:
Based on #48 this PR implements Cheerio crawler as 2nd option for crawling target web pages. To enable this feature a `useCheerioCrawler` input property and query parameter have been added. The default value for these options is false.

The Cheerio crawler works in both standby and normal mode. In standby mode the Actor will pre-create all 3 crawler (searchCrawler, playwrightContentCrawler and cheerioContentCrowler).

@metalwarrior665 @jirispilka I seem to be unable to add anyone for code review. Could you add yourself?


jirispilka (changes requested):
Thank you! A couple of minor comments below.

I think the major questions is, whether to introduce `crawlerType` instead of `useCheerioCrawler`. I would opt for  `crawlerType`. It will simplify code and is future proof

.actor/input_schema.json
```diff
@@ -136,6 +136,12 @@
             "description": "If enabled, the Actor attempts to close or remove cookie consent dialogs to improve the quality of extracted text. Note that this setting increases the latency.",
             "default": true
         },
+        "useCheerioCrawler": {
+            "title": "Use Cheerio Crawler",
+            "type": "boolean",
+            "description": "If enabled, the Actor uses the Cheerio Crawler to extract the target web page content.",
+            "default": false
+        },
```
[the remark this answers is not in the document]
matyascimbulka: I agree, switching to `crawlerType` input is more future-proof and flexible solution.

src/main.ts
```diff
@@ -75,14 +75,18 @@ if (standbyMode) {
...
-        // Pre-create default crawlers
-        log.info(`The Actor web server is listening for user requests at ${host}.`);
-        await createAndStartCrawlers(cheerioCrawlerOptions, playwrightCrawlerOptions);
+        log.info(`The Actor web server is listening for user requests at ${host}:${port}`);
+
+        // Pre-create all default crawlers
+        await createAndStartAllCrawlers(
+            searchCrawlerOptions,
+            contentCrawlerOptions,
+        );
```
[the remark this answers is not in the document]
metalwarrior665: We are now experimenting with Prettier so soon we might see more spreading in the code :/

src/crawlers.ts
```diff
@@ -118,50 +143,80 @@ async function createAndStartSearchCrawler(
...
-        ...(crawlerOptions as PlaywrightCrawlerOptions),
-        keepAlive: crawlerOptions.keepAlive,
-        requestQueue: await RequestQueue.open(key, { storageClient: client }),
-        requestHandler: async (context: PlaywrightCrawlingContext) => {
-            await requestHandlerPlaywright(context as unknown as PlaywrightCrawlingContext<PlaywrightCrawlerUserData>);
-        },
-        failedRequestHandler: ({ request }, err) => failedRequestHandlerPlaywright(request, err),
-    });
+    // Typeguard to determine if we should use Playwright or Cheerio crawler
+    const usePlaywrightCrawler = 'browserPoolOptions' in crawlerOptions;
```
metalwarrior665, line 160: The syntax 'string in object' doesn't typecheck (if you do `'garbage' in crawlerOptions;`, you will get no type warning and return false in runtime). This is bad because you rely on `browserPoolOptions` which might get removed in the future. 

You probably need to do discriminated union, this means adding `type: 'cheerio' | 'playwright'` to the `PlaywrightCrawlerOptions` and `CheerioCrawlerOptions`. You could make a wrapper type like this

```ts
type CrawlerOptions = { type: 'cheerio', crawlerOptions: CheerioCrawlerOptions } | { type: 'playwright', crawlerOptions: PlaywrightCrawlerOptions }

const fn = (options: CrawlerOptions): CheerioCrawler | PlaywrightCrawler => {
    const { type, crawlerOptions } = options;

    if (type === 'cheerio') {
        return new CheerioCrawler(crawlerOptions);
    }
    return new PlaywrightCrawler(crawlerOptions);
}
```

Another way would be to remove some layer of function call so you don't have to pass this around as union.

metalwarrior665 (changes requested):
Thanks, just minor things. Could you also share some test runs?

src/main.ts
```diff
@@ -75,14 +75,18 @@ if (standbyMode) {
...
     });
 } else {
     log.info('Actor is running in the NORMAL mode.');
     try {
-        await handleSearchNormalMode(input, cheerioCrawlerOptions, playwrightCrawlerOptions, playwrightScraperSettings);
+        await handleSearchNormalMode(input, searchCrawlerOptions, contentCrawlerOptions[0], contentScraperSettings);
```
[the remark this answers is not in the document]
metalwarrior665: This is bad, someone reordering the push can break this. I think it would be better to create distinct types for normal vs standby start so it is clear one starts cheerio | playwright while other starts both

src/crawlers.ts
```diff
@@ -26,41 +26,66 @@ export function getCrawlerKey(crawlerOptions: CheerioCrawlerOptions | Playwright
...
-    return { searchCrawler, playwrightCrawler, playwrightCrawlerKey };
+
+    return { searchCrawler, contentCrawler, contentCrawlerKey };
+}
+
+/**
+ * Creates and starts a Google search crawler and content crawlers for all provided configurations.
+ * A crawler won't be created if it already exists.
+ */
+export async function createAndStartAllCrawlers(
```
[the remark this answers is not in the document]
matyascimbulka: Yes, the difference here is in standalone vs standby mode. I'll remove the `createAndStartAllCrawlers` function and figure out how to use the `createAndStartCrawlers` for both usecases.

.actor/input_schema.json
```diff
@@ -136,6 +136,12 @@
             "description": "If enabled, the Actor attempts to close or remove cookie consent dialogs to improve the quality of extracted text. Note that this setting increases the latency.",
             "default": true
         },
+        "useCheerioCrawler": {
+            "title": "Use Cheerio Crawler",
+            "type": "boolean",
+            "description": "If enabled, the Actor uses the Cheerio Crawler to extract the target web page content.",
```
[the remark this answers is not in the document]
jirispilka: Good point with the browser vs plain HTTP, and also `scrapingTool` sounds good

.actor/input_schema.json
```diff
@@ -136,11 +136,14 @@
...
-            "type": "boolean",
-            "description": "If enabled, the Actor uses the Cheerio Crawler to extract the target web page content.",
-            "default": false
+        "scrapingTool": {
+            "title": "Which scraping tool to use",
+            "type": "string",
+            "description": "Choose what scraping tool to use for extracting the target web pages. The Browser tool is more powerful and can handle JavaScript heavy websites. While the Plain HTML tool is about two times faster.",
+            "editor": "select",
+            "default": "playwright",
+            "enum": ["playwright", "cheerio"],
```
metalwarrior665, line 145: I wanted to propose to sync these to `enumTitles` because if these are different, it creates mess with debugging and discussion later. But maybe it would be better to expand the `enumTitles` as `Browser (uses Playwright)`?

.actor/input_schema.json
```diff
@@ -136,11 +136,14 @@
...
+            "enumTitles": ["Browser", "Plain HTML"]
```
metalwarrior665, line 146: I meant `Plain HTTP` but honestly not sure what is the best term. WCC uses [Raw HTTP](https://apify.com/apify/website-content-crawler/input-schema) so maybe let's use that too, we have same/similar users.

matyascimbulka:
I have changed the values in the input schema (and corresponding constants in code). Also I have managed to suppress the experimental warning from node.

Here is a run of this version: https://console.apify.com/view/runs/ZIHv6dXe1hmlly0aX

metalwarrior665 (approved):
Great job!

src/main.ts
```diff
@@ -58,31 +58,56 @@ app.use((req, res) => {
...
-        await createAndStartCrawlers(cheerioCrawlerOptions, playwrightCrawlerOptions);
+        log.info(`The Actor web server is listening for user requests at ${host}:${port}`);
+
+        const promisese: Promise<unknown>[] = [];
+        promisese.push(createAndStartSearchCrawler(searchCrawlerOptions));
+        for (const settings of contentCrawlerOptions) {
+            promisese.push(createAndStartContentCrawler(settings));
+        }
+
+        await Promise.all(promisese);
```
jirispilka, line 91:
```suggestion
        const promises: Promise<unknown>[] = [];
        promises.push(createAndStartSearchCrawler(searchCrawlerOptions));
        for (const settings of contentCrawlerOptions) {
            promises.push(createAndStartContentCrawler(settings));
        }

        await Promise.all(promises);
```

jirispilka: Typo in promisese

jirispilka (approved):
Thank you! 

I really like it, added new functionallity and the code looks better than before 🙇🏻 
There was one typo - I added a suggestion.

I've prepared an update for readme and input_schema.json (reorder fields to move scrapingTool more to the top) in this [branch](https://github.com/apify/rag-web-browser/tree/feat/cheerio-req-handler) I can merge it once you merge this.

src/main.ts
metalwarrior665: Someone is not using [Code Spell Checker](https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker) :)

matyascimbulka: I didn't know this existed. I'm going to install it now.

matyascimbulka:
I have fixed the typo. If you're happy with it we can merge the PR. But I can't do it since I don't have the button available.

jirispilka:
> Thanks for the feedback. I have fixed the typo. If you're happy with it we can merge the PR. But I can't do it since I don't have the button available.

@matyascimbulka I thought anyone at Apify can do it. I'm not sure what is a correct setting. I added you, you should be able to merge it now.

matyascimbulka:
I have merged the PR. Do we need to build it in console or is automated?

jirispilka:
Thank you

> I have merged the PR. Do we need to build it in console or is automated?

No, we need to build in the console. Let us first merge this [PR](https://github.com/apify/rag-web-browser/pull/53) and then build it. Will you please build it and test briefly?

PR #54, opened 2025-03-11
fix: add block media feature for playwright - blocks images, videos, css

MQ37:
closes https://github.com/apify/rag-web-browser/issues/47#event-16259291157

### DIFF
--- a/.actor/input_schema.json
+++ b/.actor/input_schema.json
@@ -145,6 +145,12 @@
             "description": "If enabled, the Actor attempts to close or remove cookie consent dialogs to improve the quality of extracted text. Note that this setting increases the latency.",
             "default": true
         },
+        "blockMedia": {
+            "title": "Block media resources",
+            "type": "boolean",
+            "description": "If enabled, the Actor will block loading of images, videos and CSS resources when using the Playwright browser. This can improve performance and reduce bandwidth usage.",
+            "default": false
+        },
         "debugMode": {
             "title": "Enable debug mode",
             "type": "boolean",
--- a/src/input.ts
+++ b/src/input.ts
@@ -111,6 +111,27 @@ function createPlaywrightCrawlerOptions(input: Input, proxy: ProxyConfiguration 
                 maxConcurrency,
                 minConcurrency,
             },
+            preNavigationHooks: input.blockMedia ? [
+                async ({ page }) => {
+                    await page.route('**/*', async (route) => {
+                        const resourceType = route.request().resourceType();
+                        const url = route.request().url();
+
+                        // Block if it's an image/video/css resource type or has an image/video extension
+                        if (
+                            resourceType === 'image'
+                            || resourceType === 'video'
+                            || resourceType === 'media'
+                            || resourceType === 'stylesheet'
+                            || /\.(jpg|jpeg|png|gif|bmp|webp|mp4|webm|ogg|mov|css)$/i.test(url)
+                        ) {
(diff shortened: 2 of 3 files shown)


matyascimbulka (changes requested):
Looks great. Nice feature that will reduce proxy usage for user. Thank you.

But you've forgotten to add the input `blockMedia` into the `defaults` object in `const.ts`. https://github.com/apify/rag-web-browser/blob/a4a264184f79b6c190a6056bde2cdc8042e8229c/src/const.ts#L23-L47

This causes the input processing to drop it from the input and because of that it's never applied.

jirispilka:
@MQ37, thank you!  

I believe we want this enabled by default, right? It should be a non-breaking change that improves speed and reduces traffic with proxies.  

Could you also provide 2-3 runs with this feature enabled and disabled so we can compare perf?

MQ37:
> @MQ37, thank you!
> 
> I believe we want this enabled by default, right? It should be a non-breaking change that improves speed and reduces traffic with proxies.
> 
> Could you also provide 2-3 runs with this feature enabled and disabled so we can compare perf?

Sure, we can enable this by default :+1: 

Here is the list of runs.

Blocking disabled:
- https://console.apify.com/view/runs/SNceZzAWy0HU4PLyC
- https://console.apify.com/view/runs/awU62WFrgpfagju4y
- https://console.apify.com/view/runs/QuSuhMOvSyhHNbRsH

Blocking enabled:
- https://console.apify.com/view/runs/dDqrZHJ7DqgD2Kq1l
- https://console.apify.com/view/runs/wAGA20vFfL4rAP8oV
- https://console.apify.com/view/runs/M1NAx4X91V3bSCI4H

MQ37:
> Looks great. Nice feature that will reduce proxy usage for user. Thank you.
> 
> But you've forgotten to add the input `blockMedia` into the `defaults` object in `const.ts`.
> 
> https://github.com/apify/rag-web-browser/blob/a4a264184f79b6c190a6056bde2cdc8042e8229c/src/const.ts#L23-L47
> 
> This causes the input processing to drop it from the input and because of that it's never applied.

Thank you for noticing :+1: I tested it with local browser and it actually works without being added, but will add that :+1:

matyascimbulka (approved):
Thank you for the change. Now the Actor acts consistently between STANDALONE and STANDBY mode.

jirispilka (changes requested):
@MQ37, thank you! This looks very promising.  

I’m afraid I don’t fully understand the test and the numbers you provided.  
Did you call the Actor with several requests and then record the value?  

It seems that with each additional request, the response time gets progressively slower.  
An average response time of 45 doesn’t sound quite right.  

Also, we need to document this option and its benefits in the README 🙏🏻

MQ37:
> @MQ37, thank you! This looks very promising.
> 
> I’m afraid I don’t fully understand the test and the numbers you provided. Did you call the Actor with several requests and then record the value?
> 
> It seems that with each additional request, the response time gets progressively slower. An average response time of 47s doesn’t sound quite right.
> 
> Also, we need to document this option and its benefits in the README 🙏🏻

Added this feature to the README, thank you for noticing :+1: 

Those performance numbers are from the `src/performance-measures.ts`. Basedo on my manual testing it is about 15 - 20 % faster with the media blocking enabled.

jirispilka:
Let's keep it open. I'll merge here the other PR.

PR #59, opened 2025-03-17
fix: support all input fields as query params in standby (mainly proxy)

metalwarrior665:
I tested this locally. I think we could get rid of the Readme section about query params and just send users to https://apify.com/apify/rag-web-browser/input-schema so we don't have duplicate documentation.


src/utils.ts
```diff
@@ -3,9 +3,30 @@ import { parse, ParsedUrlQuery } from 'querystring';
...
+    const params = parse(url.slice(1));
+
+    // Parse non-primitive parameters following input schema
+    type SupportedParams = keyof typeof inputSchema.properties
+    for (const [key, value] of Object.entries(params)) {
+        // If the key is not supported by schema, skip it
+        if (!Object.keys(inputSchema.properties).includes(key)) {
+            log.warning(`Unknown parameter: '${key}', skipping it. Supported parameters: ${Object.keys(defaults).join(', ')}`);
+            continue;
+        }
```
matyascimbulka, line 18: This is a duplicate code which is already implemented in `checkAndRemoveExtraParams` in `utils.ts`. Since this code is executed first it results in duplicate warning in log.

matyascimbulka (changes requested):
I think it makes sense to support all of the input parameters. There is just a small issue with duplicate code which produces duplicate log messages.

jirispilka (changes requested):
Good point about removing query parameters from the README—it will simplify things.  

I think we should only provide examples for non-obvious ones, such as `outputFormats`. Let’s handle this in another PR. I’ve created an issue for it: #62 

I believe Matyáš meant removing the `checkAndRemoveExtraParams` function, as it is no longer needed.

metalwarrior665:
@matyascimbulka Good catch. Fixed to iterate all `defaults`.

@jirispilka I removed the `checkAndRemoveExtraParams` but the logic of parsing is now a bit more complex.

This got quite complex for such a small change (that I need for one customer) so before any further updates, I would do a larger refactor to make future updates easier. I would start with looking at the input, we could officially deprecate some fields (e.g. `keepAlive` seems redundant with Standby) and unify the types with input schema and query params, I think we can just ensure that Standby and input are the same.

jirispilka:
Ok, thank you. Yeah, it seems a bit complicated now.

jirispilka (approved):

Issue #60, opened 2025-03-18
In standby mode, a new crawler is created even though it should not be necessary.


jirispilka:
```
2025-03-18T15:50:33.520Z INFO  The Actor web server is listening for user requests at https://jiri-spilka--rag-web-browser-task.apify.actor:4321
2025-03-18T15:50:33.521Z INFO  Creating new cheerio crawler with key {"keepAlive":true,"maxRequestRetries":2,"proxyConfiguration":{"isManInTheMiddle":false,"nextCustomUrlIndex":0,"usedProxyUrls":{},"log":{"LEVELS":{"0":"OFF","1":"ERROR","2":"SOFT_FAIL","3":"WARNING","4":"INFO","5":"DEBUG","6":"PERF","OFF":0,"ERROR":1,"SOFT_FAIL":2,"WARNING":3,"INFO":4,"DEBUG":5,"PERF":6},"options":{"level":4,"maxDepth":4,"maxStringLength":2000,"prefix":"ProxyConfiguration","suffix":null,"logger":{"_events":{},"_eventsCount":0,"options":{"skipTime":true}},"data":{}},"warningsOnceLogged":{}},"domainTiers":{},"config":{"options":{},"services":{},"storageManagers":{}},"groups":["GOOGLE_SERP"],"password":"*********","hostname":"10.0.47.179","port":8011,"usesApifyProxy":true},"autoscaledPoolOptions":{"desiredConcurrency":1}}
[...]

PR #61, opened 2025-03-18
feat: performance evaluation

jirispilka:
Based on @matyascimbulka's suggestion, I refactored the code and moved `preNavigationHooks` to a separate function so that selecting `blockMedia: true/false` does not create a new instance of the crawler.  

There may be a better way to block media, but it didn’t work for me—perhaps @metalwarrior665 can help here?  

```javascript
preNavigationHooks: [
    async ({ blockRequests }) => {
        // Block all requests to URLs that include `adsbygoogle.js` and also all defaults.
        await blockRequests({
            extraUrlPatterns: ['adsbygoogle.js'],
        });
    },
],
```  

Another issue (#60) in standby mode causes multiple crawlers to be created without reason. I’ll leave this for a separate PR.  

And some number not as good as I hoped for but still it is an improvement
image


matyascimbulka (approved):
Thank you for implementing the changes. There was no need to start new crawler for blocking media.

I'm not sure why the `blockRequests` function doesn't work. But the `page.route` function seems to be the way to go for this use case (outside of Crawlee).

MQ37 (approved):
LGTM :+1: And thank you for fixing this, I haven't noticed that it spawns another crawler instance.

metalwarrior665 (changes requested):
Let's test the perf a bit more

src/crawlers.ts
```diff
@@ -130,6 +131,31 @@ export async function createAndStartContentCrawler(
     return { key, crawler };
 }
 
+/**
+ * PreNavigation hook that blocks resources based on the blockMedia setting
+ * from the request's userData.
+ * Only blocks resources if blockMedia is true.
+ */
+async function blockMediaResourcesHook({ page, request }: PlaywrightCrawlingContext<ContentCrawlerUserData>) {
+    await page.route('**/*', async (route) => {
```
metalwarrior665, line 140: `page.route` disables native browser cache which is why `blockRequests` is normally recommended (that is a native Chromium CDP call). The cache disabling is only bad if you do more requests for the same site. I would do a perf test on more URLs of the same site and test more sites because this could slow us down as well.

metalwarrior665 (changes requested):

jirispilka:
This is stale for long time, let's close it

PR #65, opened 2025-03-26
fix: return dataset it on request timeout

MQ37:
Since we cannot remove request that are already being crawled we just return dataset id on timeout so the user knows where to check for results - as we discussed.

closes https://github.com/apify/rag-web-browser/issues/31

### DIFF
--- a/src/responses.ts
+++ b/src/responses.ts
@@ -39,7 +39,8 @@ export function createResponsePromise(responseId: string, timeoutSecs: number): 
 
         // Set a timeout to reject the promise if it takes too long
         data.timeoutId = setTimeout(() => {
-            sendResponseError(responseId, 'Timed out');
+            const defaultDatasetId = process.env.ACTOR_DEFAULT_DATASET_ID;
+            sendResponseError(responseId, `Timed out. Please check the dataset ${defaultDatasetId} for results.`);
         }, timeoutSecs * 1000);
     });
 }


jirispilka:
On second thought, this goes against the idea of standby mode — it's meant to be request/response. No one will check the dataset, and that's not the goal.

It would be better to fix the issue by terminating the running request and returning empty results.
I can't image a use-case that someone would use it this way.

MQ37:
> On second thought, this goes against the idea of standby mode — it's meant to be request/response. No one will check the dataset, and that's not the goal.
> 
> It would be better to fix the issue by terminating the running request and returning empty results. I can't image a use-case that someone would use it this way.
> 
> Can you?

Agree, not many people will check the dataset and they will consume mainly the response itself. Will check if there is a simple way to skip or prevent the request from crawling.

MQ37:
@jirispilka Changed the implementation to cancel the requests of timed out response based on the discussion https://github.com/apify/crawlee/discussions/1215. This way the request content is not handled, they are only added to the dataset with status failed with reason timed out.

src/state.ts
```diff
@@ -0,0 +1,4 @@
+import { TIMED_OUT_RESPONSE_ARRAY_SIZE } from './const.js';
+import { BoundedArray } from './helpers/bouded-array.js';
+
+export const timedOutResponses = new BoundedArray<string>(TIMED_OUT_RESPONSE_ARRAY_SIZE);
```
matyascimbulka, line 4: I think this should utilize `Actor.useState()` to survive migrations. This way all of the requests that are errored out due to migration will retry after the migration completes but wont't have their id in the `BoundedArray`.

src/responses.ts
```diff
@@ -39,7 +40,8 @@ export function createResponsePromise(responseId: string, timeoutSecs: number):
 
         // Set a timeout to reject the promise if it takes too long
         data.timeoutId = setTimeout(() => {
-            sendResponseError(responseId, 'Timed out');
+            timedOutResponses.add(responseId);
```
matyascimbulka, line 43: We should probably add this to `addTimeoutToAllResponses` function.

matyascimbulka (changes requested):
I also agree that in standby mode the dataset is irrelevant. Thank you the solutions looks good. I just have small comments.

The biggest issue for me is that the bounded array doesn't survive migrations which could potentially leave some requests running. @jirispilka Do you think this could be an issue?

And there are 2 lint errors.

src/helpers/bouded-array.ts
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    delete(item: T): boolean {
+        const index = this.items.indexOf(item);
+        if (index > -1) {
+            this.items.splice(index, 1);
+            return true;
+        }
+        return false;
+    }
```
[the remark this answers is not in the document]
MQ37: currently no, but I added that so the array supports all elementary operations in case it is used in future.

matyascimbulka:
@MQ37 I'm looking at the code again and I'm wondering if we can't just check if `responseData` in `responses.ts` has the `responseId` key when handling the request in the handler. That way we wouldn't need to manage another new array and it would survive migrations.

MQ37:
> @MQ37 I'm looking at the code again and I'm wondering if we can't just check if `responseData` in `responses.ts` has the `responseId` key when handling the request in the handler. That way we wouldn't need to manage another new array and it would survive migrations.

This might be simpler but we would still need to handle the migration, right?

matyascimbulka:
I think migrations would work as expected. The `responseData` map would be initiated empty therefore any resurrected requests wouldn't find their `responseId` as its key. The missing key would signal the handler to set the `request.noRetry = true;` and throw the error.

MQ37:
> I think migrations would work as expected. The `responseData` map would be initiated empty therefore any resurrected requests wouldn't find their `responseId` as its key. The missing key would signal the handler to set the `request.noRetry = true;` and throw the error.

what actually happens to the user request to the `/search` endpoint when the Actor migrates. Does Apify act as a persist the connection that is then handed over to the new Actor container or is it closed? If it is closed we actually do not need to handle this case.

MQ37:
> > I think migrations would work as expected. The `responseData` map would be initiated empty therefore any resurrected requests wouldn't find their `responseId` as its key. The missing key would signal the handler to set the `request.noRetry = true;` and throw the error.
> 
> I guess the `responseData` approach would be simpler, thank you for suggestion 👍
> 
> What actually happens to the user request to the `/search` endpoint when the Actor migrates. Does Apify act as a persist the connection that is then handed over to the new Actor container or is it closed? If it is closed we actually do not need to handle this case.

Ahh, I see. We just sent Actor is migrating please try again and cut the connection.

src/request-handler.ts
```diff
@@ -148,7 +149,12 @@ export async function requestHandlerPlaywright(
     context: PlaywrightCrawlingContext<ContentCrawlerUserData>,
 ) {
     const { request, response, page, closeCookieModals } = context;
-    const { contentScraperSettings: settings } = request.userData;
+    const { contentScraperSettings: settings, responseId } = request.userData;
+
+    if (timedOutResponses.has(responseId)) {
+        request.noRetry = true;
+        throw new Error('Timed out. Cancelling the request...');
+    }
```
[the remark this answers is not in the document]
MQ37: moved to separate function with the refactor

src/state.ts
MQ37: refactored to use `responseData`

src/responses.ts
MQ37: refactored to use `responseData`

matyascimbulka (approved):
You're welcome. I'm happy to help. This looks good.

PR #69, opened 2025-03-31
feat: more tests, tiny refactor

MQ37:
This PR introduces basic crawler tests, standby tests, fixes linter issues, moves standby server to separate file, and adds content callback handler so we can work with extracted content in tests


src/crawlers.ts
```diff
@@ -139,9 +139,9 @@ async function createPlaywrightContentCrawler(
         ...crawlerOptions,
         keepAlive: crawlerOptions.keepAlive,
         requestQueue: await RequestQueue.open(key, { storageClient: client }),
-        requestHandler: async (context) => {
+        requestHandler: crawlerOptions.requestHandler ?? (async (context) => {
```
matyascimbulka, line 142: Is this check necessary? It seems like it would be quite easy to accidentally override the request handler and remove the intended behavior. If overriding the request handler is needed lets figure something else out.

src/crawlers.ts
```diff
@@ -155,9 +155,10 @@ async function createCheerioContentCrawler(
         ...crawlerOptions,
         keepAlive: crawlerOptions.keepAlive,
         requestQueue: await RequestQueue.open(key, { storageClient: client }),
-        requestHandler: async (context) => {
-            await requestHandlerCheerio(context as unknown as CheerioCrawlingContext<ContentCrawlerUserData>);
-        },
+        requestHandler: crawlerOptions.requestHandler ?? (async (context) => {
```
matyascimbulka, line 158: Same as `createPlaywrightContentCrawler`

src/crawlers.ts
MQ37: this is not needed any more, forgot to remove, thank you for noticing :+1:

src/request-handler.ts
```diff
@@ -107,11 +107,25 @@ async function checkValidResponse(
...
+async function pushResultToDataset(
+    input: ResultCallbackInput,
+) {
+    const { result, context, request } = input;
+    if (!context) throw new Error('Context must be passed to the pushResultToDataset function');
+    if (!request) throw new Error('Request must be passed to the pushResultToDataset function');
+
+    log.info(`Adding result to the Apify dataset, url: ${request.url}`);
+    await context.pushData(result);
+}
```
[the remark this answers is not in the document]
MQ37: Agree, it is already so complex and coupled, will use spying :+1:

MQ37:
> Thank you for cleaning up the code and adding more tests. And since you're already cleaning up the code why not upgrade `eslint` to version 9 with the new configuration (https://www.notion.so/apify/New-linting-and-formatting-configuration-across-Apify-0b62890892cd4b71b4fe2dee9291b4af?pvs=4)?
> 
> Also I have few comments about the added complexity due to the new tests. I think there might be simpler ways to achieve the desired results.

Thank you for review and suggestion :+1: Changed eslint to 9

matyascimbulka (approved):
Thank you for the changes. It looks good to me and it's nice to have some more testing.

src/types.ts
```diff
@@ -65,6 +65,17 @@ export interface TimeMeasure {
     timeDeltaPrevMs: number;
 }
 
+export interface ContentScraperSettings {
```
jirispilka, line 68: Why are you moving ContentScraperSettings.
If you need to reorder it then I would put together Input and Output, it is an obvious choice.

tests/helpers/server.ts
```diff
@@ -0,0 +1,49 @@
...
+}
+
+/**
+ * Starts a test server on the specified port
+ * @param port Port number to use
+ * @returns HTTP server instance
+ */
+export function startTestServer(port = 3030): Server {
+    const app = createTestServer();
+    const server = app.listen(port, () => {
```
jirispilka, line 28: You don't need to create server const

jirispilka (changes requested):
Love it! Thanks.

And since you've touched some parts of the code, can you provide Actors runs for a normal run and standby run? 

Just to be sure.

src/crawlers.ts
```diff
@@ -26,13 +26,40 @@ export function getCrawlerKey(crawlerOptions: CheerioCrawlerOptions | Playwright
     return JSON.stringify(crawlerOptions);
 }
 
+/**
```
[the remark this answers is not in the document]
MQ37: because of the linter - it said we are using XYZ before was defined

src/types.ts
MQ37: same here - linter said we are using interface before it was defined

tests/helpers/server.ts
MQ37: how would I otherwise close the server? It is used in `afterAll` to close the server after all tests are finished

MQ37:
> Love it! Thanks.
> 
> And since you've touched some parts of the code, can you provide Actors runs for a normal run and standby run?
> 
> Just to be sure.
> 
> I left a couple of comments. I think if you need to move a code, there should be a reason for it.

Sure, here is the list of runs :+1: 

Normal cheerio: https://console.apify.com/view/runs/CJwySHsK5FERBmClg
Normal playwright: https://console.apify.com/view/runs/tNfP6L9PhWCibcMAd
Standby (playwright + cheerio + requestTimeoutSecs test): https://console.apify.com/view/runs/qX1cszmHaKUos9scs

tests/helpers/server.ts
matyascimbulka: I think that Jirka means that you can replace the `const server =` with `return`. There is no need to save the return value of `app.listen` if you don't work with it within the function.

MQ37: ahh, I understand now :facepalm:

jirispilka (approved):
Thank you

PR #70, opened 2025-03-31
feat: unbloat, reduce image size

MQ37:
Closes https://github.com/apify/rag-web-browser/issues/72

These changes bring image size from ~1.4 GB to ~600 MB.

### DIFF
--- a/.actor/Dockerfile
+++ b/.actor/Dockerfile
@@ -8,14 +8,24 @@ COPY package*.json ./
 
 # Install all dependencies. Don't audit to speed up the installation.
 RUN npm ci
-RUN npx -y playwright install --with-deps firefox
 
-# Next, copy the source files.
+# Copy the source files.
 COPY . ./
 
 # Install all dependencies and build the project.
-# Don't audit to speed up the installation.
 RUN npm run build
+
+# Install Playwright and Firefox.
+RUN npx -y playwright install --with-deps firefox
+
+# Move the installed browser to a separate directory.
+RUN mkdir -p /app/ms-playwright \
+    && mv $(find ~/.cache/ms-playwright -type d -name "firefox" -path "*/firefox" | head -n 1) /app/ms-playwright/firefox
+
+# Move Firefox libs to a separate directory for later copy.
+RUN apt-get -y install strace && \
+    mkdir -p /firefox/lib/ && \
+    strace -e openat /app/ms-playwright/firefox/firefox 2>&1 | grep /lib/x86 | awk -F'"' '{print $2}' | xargs -I {} cp {} /firefox/lib/
 
 # Create final image
 FROM gcr.io/distroless/nodejs22-debian12
@@ -24,6 +34,12 @@ WORKDIR /app
 
 # Copy the node_modules and built app from the build stage
 COPY --from=builder /app /app
+
+# Copy Firefox libs
+COPY --from=builder /firefox/lib/* /lib/x86_64-linux-gnu/
+
+# Set the default browser path to the installed Firefox.
+ENV APIFY_DEFAULT_BROWSER_PATH=/app/ms-playwright/firefox/firefox
 
 # Run the image.
 CMD ["/app/dist/src/main.js", "--silent"]
--- a/.actor/Dockerfile
+++ b/.actor/Dockerfile
@@ -1,65 +1,29 @@
-# Specify the base Docker image. You can read more about
-# the available images at https://crawlee.dev/docs/guides/docker-images
-# You can also use any other image from Docker Hub.
-FROM apify/actor-node-playwright-chrome:22-1.46.0 AS builder
+FROM node:22-bookworm AS builder
+
+WORKDIR /app
 
 # Copy just package.json and package-lock.json
 # to speed up the build using Docker layer cache.
-COPY --chown=myuser package*.json ./
+COPY package*.json ./
 
 # Install all dependencies. Don't audit to speed up the installation.
-RUN npm install --include=dev --audit=false
+RUN npm ci
+RUN npx -y playwright install --with-deps firefox
 
-# Next, copy the source files using the user set
-# in the base image.
-COPY --chown=myuser . ./
+# Next, copy the source files.
+COPY . ./
 
 # Install all dependencies and build the project.
 # Don't audit to speed up the installation.
 RUN npm run build
 
 # Create final image
-FROM apify/actor-node-playwright-firefox:22-1.46.0
+FROM gcr.io/distroless/nodejs22-debian12
 
-# Copy just package.json and package-lock.json
-# to speed up the build using Docker layer cache.
-COPY --chown=myuser package*.json ./
+WORKDIR /app
 
-# Install NPM packages, skip optional and development dependencies to
-# keep the image small. Avoid logging too much and print the dependency
-# tree for debugging
-RUN npm --quiet set progress=false \
-    && npm install --omit=dev --omit=optional \
-    && echo "Installed NPM packages:" \
-    && (npm list --omit=dev --all || true) \
-    && echo "Node.js version:" \
-    && node --version \
-    && echo "NPM version:" \
-    && npm --version \
-    && rm -r ~/.npm
-
-# Remove the existing firefox installation
-RUN rm -rf ${PLAYWRIGHT_BROWSERS_PATH}/*
-
-# Install all required playwright dependencies for firefox
-RUN npx playwright install firefox
-# symlink the firefox binary to the root folder in order to bypass the versioning and resulting browser launch crashes.
-RUN ln -s ${PLAYWRIGHT_BROWSERS_PATH}/firefox-*/firefox/firefox ${PLAYWRIGHT_BROWSERS_PATH}/
-
-# Overrides the dynamic library used by Firefox to determine trusted root certificates with p11-kit-trust.so, which loads the system certificates.
-RUN rm $PLAYWRIGHT_BROWSERS_PATH/firefox-*/firefox/libnssckbi.so
-RUN ln -s /usr/lib/x86_64-linux-gnu/pkcs11/p11-kit-trust.so $(ls -d $PLAYWRIGHT_BROWSERS_PATH/firefox-*)/firefox/libnssckbi.so
-
-# Copy built JS files from builder image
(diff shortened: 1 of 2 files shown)


MQ37:
> @MQ37 Could you give us links to the build as well as some runs (STANDBY and STANDALONE) using it?

Sure :+1: 

Build link: https://console.apify.com/actors/B2VM9FhWyxLEMb7tm/builds/1.0.25/log
Normal run: https://console.apify.com/view/runs/osLeK8ehOOosBGuot
Standby run: https://console.apify.com/view/runs/UQmDACw95ugfnTSDm

matyascimbulka (approved):
Wow, this is really cool. Reducing the image size by 1GB without reducing functionality is impressive. Thank you.

metalwarrior665:
Have you checked if this makes sense to do for the Apify Docker base images?

MQ37:
> Have you checked if this makes sense to do for the Apify Docker base images?

Since it uses distroless image where we need to copy all the requirements I don't think it makes sense and is viable.

jirispilka:
Yeah, this is great. It will speed up start significantly (when image is not cached)

I think it could make sense to use it as a base image for certain use cases — or to include it in the templates as an example.

Assuming there isn’t some catch I’m not seeing right now?  @metalwarrior665 can you think of any?
Or discuss this with someone from tooling?

We definitely need to test it properly before releasing it in the RAG Web Browser.

metalwarrior665:
This should definitely go through tooling/platform as the base Dockerfiles are cached/preloaded in some way.

Btw: When we dropped image size in Google Maps from 2 TB to 400 MB (browser -> cheerio), we saw no improvement in startup time. I think the caching just takes care of that. So I would measure what benefit we actually want.

MQ37:
just found out that when I tested I probably tested locally with the playwright which does not use the docker image :facepalm: When including the browser and all the libs we can realistically save ~300 MB which I don't know is worth the overhead. But we can look into the distroless images for other, simpler use cases for node only Actors.

MQ37:
> just found out that when I tested I probably tested locally with the playwright which does not use the docker image and when copyting from builder image I forgot to copy the browser with it's requirements 🤦 When including the browser and all the libs we can realistically at maximum save ~300 MB which I don't know is worth the overhead (we would have to solve loading of the dynamically linked libraries). But we can look into the distroless images for other, simpler use cases for node only Actors.

Taking this back, hacked it to work (its really hacky) the image is ~600 MB and playwright works :+1: More than 300 MB but way less than original 1.3 GB.

MQ37:
Build: https://console.apify.com/actors/B2VM9FhWyxLEMb7tm/builds/1.0.37/log

**Runs**
Cheerio normal: https://console.apify.com/view/runs/ZgqKfUbfwrpJGrPlN
Playwright Firefox: https://console.apify.com/view/runs/HLqHFA2BmTWjVK5IO
Standby both: https://console.apify.com/view/runs/uHWmjNsJG1lI5A2gE

metalwarrior665:
Cool! Could you still make a post in #product-dev-tools and present this achievement? We can take this Actor as PoC of the distroless approach but it should be approved by the Tooling team. There might be some assumptions we have about our base Dockerfiles

metalwarrior665:
@MQ37 If you look at the first 2 log lines of your runs, the Docker pull takes 6+ seconds which is terrible. Usually it takes between 1-2 seconds. You can try a few more tests but I think this is cache miss and unless we make your distroless approach a new standard, we have to pause this.

jirispilka:
Yes, It would mean that the image is not cached. If you ran it more often, then the time should improve. 

Since the rag web browser is used 2-5k runs per day, we need to caution we non-standard changes.
Therefore, I'm holding with media blocking PR #54 , I need more time to properly tested it

MQ37:
> @MQ37 If you look at the first 2 log lines of your runs, the Docker pull takes 6+ seconds which is terrible. Usually it takes between 1-2 seconds. You can try a few more tests but I think this is cache miss and unless we make your distroless approach a new standard, we have to pause this.

Did a few more tests and the pull is way faster when I hit the cache. But I noticed the startup time, time from "Starting Docker container." to the printed system info in logs, is slower/unstable (higher std). Which is what we want to optimize for.

```
Original RAG Browser:
Pull to Start - Mean: 0.173s, Median: 0.179s, Std: 0.029s
Start to Sys - Mean: 1.607s, Median: 1.586s, Std: 0.177s

RAG Browser Distroless:
Pull to Start - Mean: 4.545s, Median: 4.028s, Std: 4.744s
Start to Sys - Mean: 2.123s, Median: 1.597s, Std: 0.981s

```

metalwarrior665:
We need to optimize from Actor start to system info but these parts will depend on multiple teams. 

Platform team has been working on optimizing the base Dockerfiles for years so you will definitely need to talk with them. The problem is a bit of a catch-22 as they will not want pre-cache images that are not the base for everybody. And even if we would change it for everyone, there will still be old versions and people not upgrading. But I'm just speculating, the best is to talk to them.

And then tooling team can probably influence how fast Node.js or Python process starts

MQ37:
> We need to optimize from Actor start to system info but these parts will depend on multiple teams.
> 
> Platform team has been working on optimizing the base Dockerfiles for years so you will definitely need to talk with them. The problem is a bit of a catch-22 as they will not want pre-cache images that are not the base for everybody. And even if we would change it for everyone, there will still be old versions and people not upgrading. But I'm just speculating, the best is to talk to them.
> 
> And then tooling team can probably influence how fast Node.js or Python process starts

Talked to @jirimoravcik, and the first thing is we need more test runs (thousands) to get statistically significant results. Jirka tried a few runs, and they seem okay to him - https://apify.slack.com/archives/CD0SF6KD4/p1743774919039049?thread_ts=1743622813.549859&cid=CD0SF6KD4. I already have this on my to-do list, so I will run the tests when I have time and report the results.

MQ37:
Wrote an Actor runtime benchmark and tested current `apify/rag-web-browser` master build and the distroless one from this branch. Benchmark script: https://github.com/apify/rag-web-browser/blob/a9d93437fe4698d7fd033791e29cf1a00d5cab83/benchmark/runtime.ts

Did a total of 500 runs for each Actor (master and distroless) and measured times from `Pulling Docker image of build` to `Starting Docker container` (named `pull to start time`), from `Starting Docker container` to `System info` (named `start to system time`) and then overall Actor run time in secs from the API run details.

Based on the results the distroless image performs better in `start to system time` (which is the most important in this benchmark), results in `pull to start time` are a bit worse but comparable (should be improved once deployed in production and cached) and `total run time` also improved (this is not significant since it can be network related).

**Actor run settings**:
- Memory: 1 GB
- Input:
```
{
    query: 'apify ai',
    maxResults: 1,
}
```

**Master build results**:
```
Completed 500 runs
------------------------------------------------------
Average pull to start time: 0.21 seconds
Median pull to start time: 0.17 seconds
Min pull to start time: 0.12 seconds
Max pull to start time: 4.24 seconds
Standard deviation of pull to start times: 0.20 seconds
------------------------------------------------------
Average start to system time: 1.93 seconds
Median start to system time: 1.83 seconds
Min start to system time: 1.41 seconds
Max start to system time: 4.75 seconds
Standard deviation of start to system times: 0.40 seconds
------------------------------------------------------
Average total run time: 9.62 seconds
Median total run time: 7.53 seconds
Min total run time: 4.66 seconds
Max total run time: 50.32 seconds
Standard deviation of total run times: 5.80 seconds
```

**Distroless build results**:
```
Completed 500 runs
------------------------------------------------------
Average pull to start time: 0.30 seconds
Median pull to start time: 0.18 seconds
Min pull to start time: 0.12 seconds
Max pull to start time: 10.41 seconds
Standard deviation of pull to start times: 0.82 seconds
------------------------------------------------------
Average start to system time: 1.78 seconds
Median start to system time: 1.69 seconds
Min start to system time: 1.23 seconds
Max start to system time: 5.07 seconds
Standard deviation of start to system times: 0.42 seconds
------------------------------------------------------
Average total run time: 9.20 seconds
Median total run time: 7.12 seconds
Min total run time: 4.55 seconds
Max total run time: 39.85 seconds
Standard deviation of total run times: 5.97 seconds
```

jirispilka:
@MQ37 Thanks for the detailed benchmark!

It would be great if you could summarize the results a bit more clearly, going through all those tables takes extra effort for anyone just trying to get the main takeaway :)

I don’t think we need to show all 500 runs, right? This looks like just a subset of the full 2k runs.
Also, min/max values tend to be quite noisy, using p95 or p99 would give a more meaningful picture.

That said, when looking at the data, the key metric seems to be `start to system time` and its average with standard deviation is all we need to make a decision. So it boils down to this:

| Metric                     | Master Build     | Distroless Build |
|----------------------------|------------------|------------------|
| Average start to system time | 1.90 (±0.40) sec | 1.83 (±0.38) sec |

If we wanted to be rigorous, we’d run some statistical tests like a t-test, but based on the averages and standard deviations, I don’t see a compelling reason to introduce a new image and diverge from the standard Apify base image. Sorry.

It would be nice to see the full distribution and p95, p99, but based on the code, it looks like you'd need to rerun the entire experiment again. I don't think it is worth it.

metalwarrior665:
The important part is to test these runs spaced an hour or longer so you are not hitting the same Actor run worker with already cached images. Because we pre-cache the standard Dockerfiles, master will be much more cached than any non-standard builds (try to build an Actor in Rust and watch it to take 30 seconds to start the container :) so my assumption is that the numbers will get significantly worse for Distroless (something we saw in earlier tests).

Also the image size does not seems to be a determining factor for real runs. When I did a test of Cheerio (± 400 MB) vs browser (± 2 GB), I didn't see any difference in Startup time. So it is much more to do with how our platform optimizes these than the images themselves. Our average start used to be like 5+ seconds and almost all the improvement to current <2 sec is just optimizing the platform, rather than images.

I think you make a very good case to eventually use this as a base image for most Actors. At that point we would pre-cache it and the problem with cold starts I mentioned above would not exist. But that will be a long road so you should definitely go through this with the platform & tooling teams.

MQ37:
After discussing with @jirispilka we decided it is not worth it now - maybe in the future. I will keep this as my personal TODO to discuss with the platform and tooling team.

metalwarrior665:
Yeah, this is just one Actor, but your approach could be a significant improvement for all of Apify, so let's not throw it out.

PR #73, opened 2025-05-05
fix: return suggested results and add resultType to dataset items

MQ37:
closes https://github.com/apify/rag-web-browser/issues/71


jirispilka (approved):
Thanks, looks good!

PR #84, opened 2026-05-01
fix: Final status message with scraping statistics

jirispilka:
The Actor left statusMessage empty and therefore MCP was not able to retrieve it.


jirispilka:
FYI @MQ37

PR #117, opened 2026-07-08
feat: charge multi tenant

### LINKED ISSUE #105: Update URL2M to use multi-tenant standby
I made few calls to https://url-to-markdown.apify.actor/fetch with my personal token and it hasn't shown up in my billing statement.

1. Validate that we're charging for the event when the actor is called in Standby mode.
2. Update the Readme with information how to use the actor in Standby.
   -  If possible, update the information for the Endpoints section

Example run: https://console.apify.com/organization/ZscMwFR5H7eCtWtyh/actors/oK4fp52ayRqNayFTw/runs/AQeYKRL4sgS4fYeT2#log

Multi-tenant docs:
https://app.notion.com/p/apify/Multi-tenant-standby-developer-facing-documentation-38af39950a22806cb407e6f0bb825864?source=copy_link

ruocco-l:
Closes #105 

Tested manually with cloned actor.

### DIFF
--- a/src/crawlers.ts
+++ b/src/crawlers.ts
@@ -276,7 +276,7 @@ async function chargeStandby(eventName: string, userAuthorization: string): Prom
         log.warning(`Skipping standby charge for ${eventName} event: missing apiBaseUrl/actorRunId/token from Actor.getEnv().`);
         return;
     }
-    const url = `${apiBaseUrl}/v2/actor-runs/${actorRunId}/charge`;
+    const url = `${apiBaseUrl}v2/actor-runs/${actorRunId}/charge`;
     const response = await fetch(url, {
         method: 'POST',
         headers: {
--- a/src/crawlers.ts
+++ b/src/crawlers.ts
@@ -21,7 +21,7 @@ import { getMiniActor } from './mini-actors.js';
 import { failedRequestHandler, requestHandlerCheerio, requestHandlerPlaywright } from './request-handler.js';
 import { addEmptyResultToResponse, sendResponseError } from './responses.js';
 import type { ContentCrawlerOptions, ContentCrawlerUserData, SearchCrawlerUserData } from './types.js';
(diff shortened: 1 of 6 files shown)


nikitachapovskii-dev:
are you sure slash is redundant? https://docs.apify.com/sdk/js/reference/3.0/class/Configuration
Screenshot 2026-07-08 at 16 19 21

ruocco-l:
> are you sure slash is redundant? 

Yes, I had a failing run on a missing page. I'll point it out the discrepancy on Slack

Screenshot 2026-07-08 at 15 34 59

src/crawlers.ts
```diff
@@ -241,23 +243,73 @@ async function createCheerioContentCrawler(
...
+        return;
+    }
+    const eventName = getEventName(crawlerType);
     try {
-        if (getMiniActor().name === 'url-to-markdown') {
-            await Actor.charge({ eventName });
+        if (isActorStandby()) {
+            if (!userAuthorization) {
+                log.warning(`Skipping standby charge for ${eventName} event: missing userAuthorization (x-apify-user-authorization header was not provided).`);
+                return;
```
nikitachapovskii-dev, line 308: is authorization provided always on platform?
Can't find it in notion docs 🤔 
Can we do chargeNormal in case `if (!userAuthorization)` instead of skipping?

ruocco-l: It would not charge if we just charge from the sdk, that was the original problem from the issue

ruocco-l: The early return and the log is there to exactly show us that we do not have this piece of information, instead of just trying to call the endpoint without knowing if we have it or not, risking an obscure error response

nikitachapovskii-dev (approved):
No further questions, thanks for the solid work!

nicklamonov:
Can I test it also on a clone actor (or better some `beta` build)?
Or if there is low risk, feel free to merge it and I'll check on the actual actor.

PR #118, opened 2026-07-16
fix: guard against invalid hostname

### LINKED ISSUE #88: Propagate invalid hostname error instead of returning 200 with empty content
## Summary

When the underlying crawl fails (DNS resolution, proxy refused, timeout, invalid hostname, etc.), the standby `/search` endpoint still responds with **HTTP 200** and a result item containing metadata but no `markdown`. This makes failures indistinguishable from "page genuinely had no text content" and forces every wrapper around RAG to infer error state by checking `markdown === ''`, which is brittle and conflates two very different outcomes.

## Reproduction

Send a request with a structurally-valid but unresolvable URL:

```
GET https://rag-web-browser.apify.actor/search?query=https%3A%2F%2Fapifycom&outputFormats=markdown&scrapingTool=raw-http&maxResults=1
Authorization: Bearer <APIFY_TOKEN>
```

### What happens in RAG's logs

```
INFO  Skipping Google Search query as "https://apifycom" is a valid URL
INFO  Added request to the cheerio-content-crawler: https://apifycom/
WARN  CheerioCrawler: Reclaiming failed request back to the list or queue. Detected a session error, rotating session...
Proxy responded with 400 Bad Request: 60 bytes.
Below is the first 100 bytes of the proxy response body:
Invalid hostname: provided hostname "apifycom" is not valid.
... (10 retries, same error each time) ...
ERROR CheerioCrawler: Request failed and reached maximum retries.
INFO  Adding result to the Apify dataset, url: https://apifycom/
INFO  Response Jr4Bfwu5Ex resolved successfully with 1 results.
```

### What the caller sees

`HTTP 200 OK`, body is a JSON array with one item whose `markdown` is missing/empty. None of the diagnostic information (proxy 400, "Invalid hostname", retry count, error class) reaches the caller.

## Why this matters

Real upstream wrappers (e.g. `apify/page-scraper`, MCP clients, agent harnesses) need to:

- Tell users *why* a URL failed so they can fix typos / pick a different scraping tool / report a bug.
- Skip dataset writes / billing for failed URLs.
- Trigger fallback logic (e.g. retry with a different `scrapingTool`).

Today all of that has to be done by inspecting `markdown === ''`, which:
- Cannot distinguish "page failed to load" from "page loaded but had no extractable text" (e.g. an image-only page, a 404 page with an empty body).
- Loses every detail of *why* the crawl failed.

### 1. Return non-2xx HTTP status when the underlying crawl fails

When the crawler exhausts retries:

- DNS / hostname / connection refused → `502 Bad Gateway`
- Timeout → `504 Gateway Timeout`
- Page returned HTTP 4xx → forward that 4xx (or wrap as `502`)
- Anti-bot block → `403` or `502`

Body can still be JSON with diagnostic details. This is the standard HTTP semantic and the cheapest fix for callers.

### 2. (Or, if keeping the 200 envelope) Include an explicit `error` field per result item

```json
[{
  "metadata": { "url": "https://apifycom/" },
  "crawl": {
    "httpStatusCode": null,
    "loadedAt": "2026-05-28T12:45:23.130Z"
  },
  "error": {
    "code": "INVALID_HOSTNAME",
    "message": "Proxy responded 400: Invalid hostname: provided hostname \"apifycom\" is not valid.",
    "upstreamStatusCode": 400,
    "retryCount": 10
  }
}]
```

Callers can branch on \`error\` being present and surface \`error.code\`/\`error.message\` to their users.

### 3. Surface proxy/upstream error codes in \`crawl\`

Even without a separate \`error\` field, populating \`crawl.httpStatusCode\` and a new \`crawl.errorMessage\` from the underlying CheerioCrawler/PlaywrightCrawler failure would be a strict improvement. Today \`crawl\` is essentially empty when the crawl fails.

## Recommended path

Combine **#1** and **#3**: return `502`/`504` for hard failures, and always populate `crawl.httpStatusCode` / `crawl.errorMessage` so callers using either pattern get actionable info.

## Related code

- `src/server.ts` — `handleSearchRequest` is where the response is built.
- `src/const.ts` — `Routes.SEARCH`.
- `src/crawlers/*` — where the CheerioCrawler / PlaywrightCrawler errors are caught today.

## Context

Without this signal, wrappers can't reliably tell their users why a URL failed, and we have to assume any empty-markdown response is a failure — which would also incorrectly drop legitimate empty pages.

ruocco-l:
Closes #88 

Query and url where not checked at all at input, letting the scraper fail when trying to do something with it. This was done to avoid the standby failing at startup.

Now the checks are moved inside the search/fetch workflow and the input checks only what kind of crawlers it has to warmup.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -238,7 +238,9 @@ export function interpretAsUrl(input: string): string | null {
 
         try {
             const url = new URL(trimmed);
-            if (/^https?:/i.test(url.protocol)) return url.href;
+            const hasValidProtocol = /^https?:/i.test(url.protocol);
+            const hasValidHostname = url.hostname.split('.').length > 1 || url.hostname === 'localhost';
+            if (hasValidProtocol && hasValidHostname) return url.href;
         } catch {
             // Continue by checking whether this is a supported URL without a protocol.
         }
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -239,7 +239,7 @@ export function interpretAsUrl(input: string): string | null {
         try {
             const url = new URL(trimmed);
             const hasValidProtocol = /^https?:/i.test(url.protocol);
-            const hasValidHostname = url.hostname.split('.').length > 1 || url.hostname === 'localhost';
+            const hasValidHostname = url.hostname.split('.').filter(Boolean).length > 1 || url.hostname === 'localhost';
             if (hasValidProtocol && hasValidHostname) return url.href;
         } catch {
             // Continue by checking whether this is a supported URL without a protocol.
--- a/src/input.ts
+++ b/src/input.ts
@@ -18,7 +18,6 @@ import type {
     SERPProxyGroup,
     UrlToMarkdownInput,
 } from './types.js';
-import { interpretAsUrl } from './utils.js';
 
 /**
  * Processes the input and returns an array of crawler settings. This is ideal for startup of STANDBY mode
@@ -101,10 +100,9 @@ async function processRagWebBrowserInput(input: Partial<RagWebBrowserInput>, sta
     }> {
     /* eslint-disable no-param-reassign */
 
-    // Throw an error if the query and is not provided and standbyInit is false.
-    if (!input.query && !standbyInit) {
-        throw new UserInputError('The `query` parameter must be provided and non-empty.');
-    }
+    // Note: `query` is intentionally not validated here. It is a per-request property (depending on
+    // the mini-actor) rather than a startup/crawler-configuration property, so its presence and validity
+    // are checked when the request is actually formed (see `prepareRequest` in search.ts).
 
     // Max results
     input.maxResults = validateRange(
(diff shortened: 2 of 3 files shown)


src/utils.ts
```diff
@@ -238,7 +238,9 @@ export function interpretAsUrl(input: string): string | null {
 
         try {
             const url = new URL(trimmed);
-            if (/^https?:/i.test(url.protocol)) return url.href;
+            const hasValidProtocol = /^https?:/i.test(url.protocol);
+            const hasValidHostname = url.hostname.split('.').length > 1 || url.hostname === 'localhost';
```
nikitachapovskii-dev: hasValidHostname is in the shared interpretAsUrl. If I understand it correctly from discussion in issue, this should be only on /fetch, but not on /search

Also, сlaude spotted an edge case: a trailing dot bypasses the check, interpretAsUrl('https://apifycom.') returns https://apifycom./, so it still reaches the crawler and reproduces the empty-200 case.

nikitachapovskii-dev (commented):
Pls check comment below. BTW I think we may get back on that topic somewhen in future, so we should also cover it with couple of tests as well, wdyt?

src/utils.ts
ruocco-l: > a trailing dot bypasses the check

ah, that's a good catch, thanks.

ruocco-l: > hasValidHostname is in the shared interpretAsUrl. If I understand it correctly from discussion in issue, this should be only on /fetch, but not on /search

Hmm that's true but I think that we have to rethink this and accept that this is correct. If you run `/search?quer=https://apifycom` it will not pass the url validation and it will try to perform a search on Google. Not the intended behavior of course, but a coherent fallback.

What's your opinion on this @nicklamonov ?

nicklamonov: > If you run /search?quer=https://apifycom it will not pass the url validation and it will try to perform a search on Google. Not the intended behavior of course, but a coherent fallback.

Sounds ok to me for the /search endpoint. Didn't we have it the same up to now as well?

ruocco-l: > Didn't we have it the same up to now as well?

Yes, it's the same for similar misspells like `https://apify com` (with a space)

nikitachapovskii-dev (approved):
🚀

nicklamonov (approved):
Looks good. Thanks!

PR #120, opened 2026-07-21
chore(docs): add openAPI schema

### LINKED ISSUE #92: RAG: Update standby server OpenAPI schema
Display information how to send GET requests to standby endpoints

ruocco-l:
Closes #92 

Adding openAPI for both RAG and URL2M

### DIFF
--- a/docs/url-to-markdown-standby-openapi-3.1.0.json
+++ b/docs/url-to-markdown-standby-openapi-3.1.0.json
@@ -15,8 +15,8 @@
             "get": {
                 "operationId": "apify_url-to-markdown",
                 "x-openai-isConsequential": false,
-                "description": "An Apify Actor that takes a URL as input and returns the content of the page in Markdown format.",
-                "summary": "An Apify Actor that takes a URL as input and returns the content of the page in Markdown format.",
+                "description": "Fetches the web page at the given URL and returns its content converted to Markdown.",
+                "summary": "Fetches the web page at the given URL and returns its content converted to Markdown.",
                 "parameters": [
                     {
                         "name": "url",
--- a/docs/url-to-markdown-standby-openapi-3.1.0.json
+++ b/docs/url-to-markdown-standby-openapi-3.1.0.json
@@ -1,0 +1,65 @@
+{
+    "openapi": "3.1.0",
+    "info": {
+        "title": "URL to Markdown",
+        "description": "An Apify Actor that takes a URL as input and returns the content of the page in Markdown format.",
+        "version": "v1"
+    },
+    "servers": [
+        {
+            "url": "https://url-to-markdown.apify.actor"
+        }
+    ],
+    "paths": {
+        "/fetch": {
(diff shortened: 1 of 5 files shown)


nicklamonov:
As commented in Slack:

URL2M:
- the description of the endpoint says "An Apify Actor that takes....". Should we rather keep such actor description at the actor level and change the description for the endpoint a bit? Functionally it's fine.

RAG 
- I also don't like a part of description. Maybe we should remove this - "Web browser for OpenAI Assistants API and RAG pipelines, similar to a web browser in ChatGPT."? 
- At the same I would mention that if only the URL is provided, we do a fetch of that URL only. Now it's not explicit and humans/agents might not understand the difference.

nicklamonov (approved):
Thanks! All good from my side.

docs/url-to-markdown-standby-openapi-3.1.0.json
nikitachapovskii-dev, line 1: Did you skip some fields because they are hidden? (removeElementsCssSelector, desiredConcurrency, debugMode)
Should we in that case skip requestTimeoutSecs and debugMode for rag?

docs/url-to-markdown-standby-openapi-3.1.0.json
```diff
@@ -0,0 +1,65 @@
...
+                            "default": "raw-http"
+                        }
+                    },
+                    {
+                        "name": "proxyConfiguration",
+                        "in": "query",
+                        "description": "Apify Proxy configuration used for scraping the target web pages.",
+                        "required": false,
+                        "schema": {
+                            "type": "object",
```
nikitachapovskii-dev, line 50: I asked Claude to compare this PR with the web-fetch one ([apify/actor-web-fetch#1](https://github.com/apify/actor-web-fetch/pull/1)) and it flagged a difference that proxyConfiguration here has type: object while there they got headers type: string (JSON-encoded) for the GET
I guess we should align it here to be string as well so it JSON.parse's correclty

```
                    {
                        "name": "headers",
                        "in": "query",
                        "required": false,
                        "schema": {
                            "type": "string",
                            "description": "Additional HTTP headers to send with the request, as a JSON-encoded object string."
                        },
                        "example": "{\"Accept-Language\":\"de-DE\"}"
                    },
```

nikitachapovskii-dev (commented):
Thank you, @ruocco-l !

I noticed rag readme is updated for openapi but url2m not, maybe we should update it as well?

docs/url-to-markdown-standby-openapi-3.1.0.json
ruocco-l: Yeah, this is correct, I'll change it

docs/url-to-markdown-standby-openapi-3.1.0.json
ruocco-l: Yeah we can do that

Issue #123, opened 2026-07-23
Track HTTP requests on standby actors


ruocco-l:
Keep track of the HTTP request and make a decision on keep supporting `/sse` and `/message`.
Also, track custom crawler parameter on `/search` and `/fetch` to decide on default options.

PR #124, opened 2026-07-23
chore: create a log for http requests

ruocco-l:
Related to #123 
Once this is merged I will add the secrets

### DIFF
--- a/src/request-logger.ts
+++ b/src/request-logger.ts
@@ -16,7 +16,7 @@ const REQUIRED_QUERY_PARAM: Record<string, string> = {
 /**
  * Initialize request logging (no-op if REQUEST_LOG_DATASET_ID is not set). Call once at startup.
  */
-export async function initRequestLogger(): Promise<void> {
+export function initRequestLogger(): void {
     const datasetId = process.env.REQUEST_LOG_DATASET_ID;
     if (!datasetId) return;
 
@@ -44,15 +44,14 @@ export function logRequest(pathname: string, query: Record<string, unknown>): vo
         const requiredParamKey = REQUIRED_QUERY_PARAM[pathname];
         let loggedQuery: Record<string, string> = {};
 
+        delete loggedQuery.token;
+
         if (requiredParamKey !== undefined) {
             const keys = Object.keys(query).filter((key) => key !== requiredParamKey);
             if (keys.length === 0) return;
             loggedQuery = { ...(query as Record<string, string>) };
             delete loggedQuery[requiredParamKey];
         }
-
-        // Never persist the token query param (used for standby-mode auth) into the logged item.
-        delete loggedQuery.token;
 
         const item = {
             miniActor: getMiniActor().name,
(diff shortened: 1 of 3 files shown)


src/request-logger.ts
```diff
@@ -0,0 +1,67 @@
...
+
+        if (requiredParamKey !== undefined) {
+            const keys = Object.keys(query).filter((key) => key !== requiredParamKey);
+            if (keys.length === 0) return;
+            loggedQuery = { ...(query as Record<string, string>) };
+            delete loggedQuery[requiredParamKey];
+        }
+
+        // Never persist the token query param (used for standby-mode auth) into the logged item.
+        delete loggedQuery.token;
```
nikitachapovskii-dev: if we have token in query then we should remove it/ignore before we check for custom fields.

src/request-logger.ts
```diff
@@ -0,0 +1,67 @@
...
+// For these routes, the single allow-listed key alone doesn't warrant logging (see logRequest).
+const REQUIRED_QUERY_PARAM: Record<string, string> = {
+    [Routes.SEARCH]: 'query',
+    [Routes.FETCH]: 'url',
+};
+
+/**
+ * Initialize request logging (no-op if REQUEST_LOG_DATASET_ID is not set). Call once at startup.
+ */
+export async function initRequestLogger(): Promise<void> {
```
nikitachapovskii-dev: nit: it can be sync

nikitachapovskii-dev (commented):
Thanks, @ruocco-l !
Please check comments below. Otherwise looks good

nicklamonov:
I don't think I can review it properly until it's on the platform. ) Ping me please, when it's in a `beta` build, for example.

src/request-logger.ts
```diff
@@ -0,0 +1,66 @@
...
+    if (!enabled || !dataset) return;
+
+    try {
+        const requiredParamKey = REQUIRED_QUERY_PARAM[pathname];
+        let loggedQuery: Record<string, string> = {};
+
+        delete loggedQuery.token;
+
+        if (requiredParamKey !== undefined) {
+            const keys = Object.keys(query).filter((key) => key !== requiredParamKey);
```
nikitachapovskii-dev:
```suggestion
            const keys = Object.keys(query).filter((key) => key !== requiredParamKey && key !== 'token');
```

nikitachapovskii-dev: and move `delete loggedQuery.token;` right under `delete loggedQuery[requiredParamKey];` so it's not after declaration 😄

nikitachapovskii-dev (commented):
hmm, as it is now the token still should end up in the log and slip past the empty check 🤔 
Would something like this work?

nikitachapovskii-dev (approved):
no blockings, thank you!

nicklamonov (approved):
I'll check it when it's on the platform. 
Approving for now.

PR #125, opened 2026-07-24
fix: fall back to cleaned HTML when Readability extracts no content

### LINKED ISSUE #108: Switch from readableText to readableTextIfPossible
Introduce a fallback to `none` if Readability fails.

**Before implementation:**
Validate how many runs with Readability have not been successful in the HTML transformation.

**During implementation:**
Validate performance impact.

nikitachapovskii-dev:
follow up to https://github.com/apify/actor-rag-web-browser/pull/119

more bulletproof addition to fallbackToNone: true. If readability completely failed -> process raw html

Closes #108

### DIFF
--- a/src/website-content-crawler/text-extractor.ts
+++ b/src/website-content-crawler/text-extractor.ts
@@ -36,6 +36,10 @@ export async function readableText({
 
     const readabilityRoot = parsed?.content as HTMLElement | null;
 
+    if (options?.fallbackToNone && !readabilityRoot?.textContent?.trim()) {
+        return html;
+    }
+
     if (readabilityRoot && parsed?.title) {
         const titleElement = dom.window.document.createElement('h1');
         titleElement.textContent = parsed.title;


ruocco-l (approved):

nikitachapovskii-dev:
@nicklamonov do we want to merge it? This is related to discussion about readability we had previous week

nicklamonov:
I don't think that users would like to receive HTML when they expect Markdown.
So I would rather return empty response (or maybe even an error with a clear indication that we couldn't get the content), not raw HTML.

nikitachapovskii-dev:
It won't be raw HTML for user. readableText() returns HTML by definition (it's the HTML-transformer stage), and its result always goes through Turndown, so the user still gets Markdown, just not pruned by Readability. 

Example:
html
```
<div id="panel" style="display:none">
  <h2>Retry policy</h2>
  <p>Each request is retried up to <b>3</b> times. See <a href="https://acme.dev/docs/errors">error codes</a>.</p>
  <ul><li>Backoff: 1s, 4s, 16s</li><li>Timeout: 30s</li></ul>
</div>
```
markdown in the output:

```
## Retry policy

Each request is retried up to **3** times. See [error codes](https://acme.dev/docs/errors).

*   Backoff: 1s, 4s, 16s
*   Timeout: 30s
```
Also note removeElementsCssSelector already strips nav, footer, script, style, noscript, svg before this.



https://console.apify.com/admin/users/ZscMwFR5H7eCtWtyh/actors/runs/Pw8wvItEnLo4qCGRJ#output

nicklamonov:
Thanks for the details!
In that case I'm fine with it.

PR #127, opened 2026-07-29
perf: don't download media files

### LINKED ISSUE #109: Ignore media files
Drop media files, so we save on bandwidth.

nikitachapovskii-dev:
Closes #109.

This went bigger than I expected

- `createRequest` marks media-file URLs with `skipNavigation`; both request handlers return an empty result with `Skipped media file` instead of parsing a page.
- New `src/media.ts`: `isMediaUrl` (extension of the last path segment) and `blockMediaRequests`, a `page.route` handler aborting `image`/`media`/`font`.
- The blocker is registered in the Playwright pre-navigation hook and again after `enableBlockingInPage`, because the Ghostery handler continues everything it doesn't block.
- Pulled the skipped-result building out of `checkValidResponse` into `pushSkippedResult`, shared by both call sites. It now takes the status code from the caller — the old `response.status()` doesn't exist on the raw-http response.
- Docs in both Actor READMEs, unit tests for `isMediaUrl`, three standby tests using a request counter on the test image.

Applied to both url2m and rag

### DIFF
--- a/actors/apify_rag-web-browser/README.md
+++ b/actors/apify_rag-web-browser/README.md
@@ -224,6 +224,15 @@ Here are specific situations that might occur when the timeout is reached:
   => the Actor extracts content from the currently loaded HTML
 
 
+### Media files
+
+Media files carry no text for the LLM, so the Actor never downloads them:
+
+- Images, audio, video, and fonts of the scraped page are blocked in Browser mode (`scrapingTool=browser-playwright`).
+  This saves bandwidth and often speeds up the page load, and it has no effect on the extracted content.
+- Search results (and a `query` that is a URL) pointing directly to a media file, e.g. `https://example.com/video.mp4`,
+  are not crawled at all. Such a result is returned with an empty text and `Skipped media file` as the HTTP status message.
+
 ### Reducing response time
 
 For low-latency applications, it's recommended to run the RAG Web Browser in Standby mode
--- a/actors/apify_url-to-markdown/README.md
+++ b/actors/apify_url-to-markdown/README.md
@@ -83,6 +83,9 @@ Markdown is the perfect format to feed large language model (LLM). It is a less 
 Using markdown instead of html can help you lower the AI token cost.
 
 
+### What happens with media files?
+Media files carry no text to convert, so they are never downloaded. Images, audio, video, and fonts of the converted page are blocked in the Browser mode, and a URL pointing directly to a media file, e.g. `https://example.com/video.mp4`, is not fetched at all — the output contains no content and `Skipped media file` as the HTTP status message.
+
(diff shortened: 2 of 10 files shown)


nikitachapovskii-dev:
Scope question before I finalise this: should we also drop URLs that *are* media files
(e.g. `query=https://example.com/photo.png`), or only block images/video/fonts loaded by the page?

also should we mention this in readme?
@nicklamonov, @ruocco-l

nicklamonov:
<h2>Headline</h2>

Metric | Requests | % of all
-- | -- | --
All requests | 2,537,469 | 100%
Search terms | 1,280,564 | 50.47%
URLs | 1,256,905 | 49.53%
Media-file URLs | 8,641 | 0.34%
— of which images/video/audio only | 1,536 | 0.06%

<h2>Breakdown</h2>

Category | Requests | % of all | Top extensions
-- | -- | -- | --
Documents | 7,015 | 0.277% | pdf 6,826 · xlsx 112 · docx 39 · doc 18 · xls 14
Images | 1,512 | 0.060% | jpg 746 · png 381 · svg 202 · avif 77 · webp 49 · jpeg 46
Archives/binaries | 90 | 0.004% | exe 22 · gz 20 · zip 17 · deb 10 · dmg 10
Video | 16 | 0.0006% | mp4 15 · m3u8 1
Audio | 8 | 0.0003% | ogg 3 · mp3 3 · m4a 2

The `.mp4`/`.jpg` case from your example is far rarer than PDFs — PDF alone is 79% of all media hits. Actual video is 16 requests in 11 weeks.

Two adjacent buckets not in the 8,641, in case you want them counted:

- 596 requests (0.024%) reference a media file inside the URL's query string rather than the path (e.g. ...?outImageAddress=https%3A%2F%2F...png). Upper bound including these: ~9,237 (0.36%).
- 54 requests are search phrases (with spaces) ending in a media extension — not URLs.

<h2>Trend</h2>
Media share is growing: 0.14% in mid-May → 0.46–0.48% in late July (~3x), with absolute volume up from ~230 to ~1,450 per week. Small in share, but the direction is consistent.


<h2>Summary</h2>
Before I read the "Trend" part I was thinking to drop media requests. But then it got me thinking that maybe we shouldn't.

At the same time, we can't really convert an image to markdown. So I would probably block media even if it's a URL (not docs though).

If it's not a one-liner - let's skip it for now. If it is, let's drop media support for URLs too (images and videos only for now).

nikitachapovskii-dev:
Great, that's exactly what the PR does + skipping audio
Documents are untouched.

nicklamonov (approved):
I will test on some `beta` version, when it's released. So approving for now.

ruocco-l (approved):
Good job with this! 🚀

PR #128, opened 2026-07-29
chore: prefer <head> title, ignore SVG and iframe titles

### LINKED ISSUE #110: Update title extraction
Title extraction is implemented better in WCC because RAG uses a simple `$('title').first().text()` style extraction. WCC has an extractTitle() helper that prioritizes head title and avoids false titles inside SVGs or Crawlee iframe replacement nodes. 

**During implementation:**
Validate for possible performance degradation

nikitachapovskii-dev:
Closes #110.

- Added the extractTitle helper from wcc: it takes the title from the head of the page and ignores titles inside SVG icons and inside Crawlee iframe placeholders.
- Used it for the title in the output metadata and for the title passed to the transformer.
- Tests: four unit tests for the helper, plus a title check in the existing Cheerio crawler test.

### DIFF
--- a/tests/html-processing.test.ts
+++ b/tests/html-processing.test.ts
@@ -1,0 +1,47 @@
+import { load } from 'cheerio';
+import type { CheerioAPI } from 'crawlee';
+import { describe, expect, it } from 'vitest';
+
+import { extractTitle } from '../src/website-content-crawler/html-processing.js';
+
+// The `cheerio` version bundled with Crawlee differs from the top-level one, so the types don't match.
+const parse = (html: string) => load(html) as unknown as CheerioAPI;
+
+describe('extractTitle', () => {
+    it('should extract the title from somewhere else if not in head', () => {
+        const $ = parse(`<html>
+            <head></head>
+            <body>
+                <div class="content">The part with the content.</div>
+                <title>Title in body</title>
+            </body>
+        </html>`);
+        expect(extractTitle($)).toBe('Title in body');
+    });
+
+    it('should ignore titles in SVGs anywhere in the html', () => {
+        const $ = parse(`<html>
+            <head><svg><title>Title in head svg</title></svg></head>
+            <body>
+                <div class="content">The part with the content.</div>
+                <svg><title>Title in body svg</title></svg>
+            </body>
+        </html>`);
+        expect(extractTitle($)).toBe('');
+    });
+
+    it('should ignore titles in .crawlee-iframe-replacement anywhere in the html', () => {
+        const $ = parse(`<html>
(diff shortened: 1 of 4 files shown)


nikitachapovskii-dev:
## Performance validation

Benchmarked `extractTitle($)` vs the old `$('title').first().text()` on an already-parsed `$`
(7 rounds × N iterations, median), in both parser modes — htmlparser2 (`CheerioCrawler`) and
parse5 (Playwright `parseWithCheerio()`). Numbers below are htmlparser2; parse5 matched it.

| Case | old | new | ratio |
|---|---:|---:|---:|
| 4 real pages (Wikipedia, GitHub, Apify Store, MDN — 181–508 KB) | 0.125–0.161 ms | 0.128–0.171 ms | 1.0–1.1× |
| Real page with `<title>` stripped (2nd selector runs) | 0.145 ms | 0.291 ms | 2.0× |
| Synthetic 36k elements, 2,000 SVG titles, no title anywhere | 2.20 ms | 5.08 ms | 2.3× |

It's called twice per page, so: **+0.01 ms on real pages** (0.02–0.04% of 14–34 ms of local
processing), **+4.9 ms worst case** (2% of 245 ms). The 2.3× only shows up when both selectors
miss and each walks the whole DOM i.e. only on pages with no `<title>` at all.

сс @nicklamonov

tests/html-processing.test.ts
```diff
@@ -0,0 +1,47 @@
+import { load } from 'cheerio';
+import type { CheerioAPI } from 'crawlee';
+import { describe, expect, it } from 'vitest';
+
+import { extractTitle } from '../src/website-content-crawler/html-processing.js';
+
+// The `cheerio` version bundled with Crawlee differs from the top-level one, so the types don't match.
+const parse = (html: string) => load(html) as unknown as CheerioAPI;
```
ruocco-l, line 8: This is an annoying cast, I think we can solve it by playing with Cheerio versions, I think it was discussed some time ago

ruocco-l (approved):
Just a comment on the casting that doesn't really matter to the PR, just wanted to point it out

nicklamonov:
- In performance table are we really talking about milliseconds or seconds? If really milliseconds, then spending just a couple of extra milliseconds per title is totally not a big deal. :) 
- Isn't the title of the PR confusing (because there were more changes done)?

nikitachapovskii-dev:
> In performance table are we really talking about milliseconds or seconds? 

yes, real miliseconds, agree.
updated the pr title

nicklamonov:
I still don't see PR title update )

nikitachapovskii-dev:
> I still don't see PR title update )

damn, it wasn't saved. Re-updated now

PR #131, opened 2026-08-03
feat: expand clickable elements

### LINKED ISSUE #112: Expand clickable elements (in browser mode)
Currently RAG doesn't expand dynamic elements, so some content might be missing.

Recheck performance impact.

Risk:
Possibly an update of the navigation logic is needed.

nikitachapovskii-dev:
In Browser mode the Actor now clicks the collapsed elements (`[aria-expanded="false"]`) before extracting the content, same as WCC. No new input field, and the Raw HTTP mode is unchanged.

Measured on 12 real pages: clicking costs about 0.5 s on pages that have collapsed elements and ~7 ms on pages that don't. 

Closes #112


src/request-handler.ts
```diff
@@ -227,6 +260,11 @@ export async function requestHandlerPlaywright(
         addTimeMeasureEvent(request.userData, 'playwright-remove-cookie');
     }
 
+    if (page) {
+        await expandClickableElements(page, CLICK_ELEMENTS_CSS_SELECTOR);
+        addTimeMeasureEvent(request.userData, 'playwright-expand-clickable-elements');
+    }
+
```
ruocco-l, line 267: If you move this up, before the wait for dynamic content function, you don't have to add a synthetic wait

actors/apify_rag-web-browser/README.md
```diff
@@ -234,6 +235,19 @@ Media files carry no text for the LLM, so the Actor never downloads them:
...
+FAQ sections, and add it to the page only when it's clicked. To capture such content, the Actor clicks
+the collapsed elements of the page, i.e. those matching the `[aria-expanded="false"]` CSS selector,
+in Browser mode (`scrapingTool=browser-playwright`) before it extracts the content.
+
+Elements linking to another page are not clicked, so that the Actor stays on the page it was asked
+to extract.
+
+Note that clicking costs about half a second on pages that have such elements, and that content of
+expanded navigation menus can end up in the output as well.
+
```
ruocco-l: They don't have a way to set it off, so there is no use in saying this

ruocco-l (approved):
I will be pre-approving because I think it's correct, but let's all agree that this is a no-opt, non-customizable feature for the user. Some pages (it happened in WCC but I can't seem to find the issue related to it) may be broken by clicking (because they cause navigation) de facto making it impossible to scrape them with headless browser on RAG

src/request-handler.ts
nikitachapovskii-dev: Rechecked with scripts to measure: on real pages (3 runs each): clicking before the wait loses content: openai.com −785 chars, github.com −422

Before the wait the page usually isn't hydrated (we go with `domcontentloaded`), so the collapsed elements have no click listeners yet, or don't exist at all.

PR #138, opened 2026-08-17
chore: use x-actor-request-id to charge

ruocco-l:
Just changing the header we use on charging to be up to specs and its related function/variable names.

We had to move around the `sendResponseIfFinished` because [we want to charge before sending the request](https://apify.slack.com/archives/C09JK2NUG57/p1786974841585969?thread_ts=1786967383.595349&cid=C09JK2NUG57).

### DIFF
--- a/src/crawlers.ts
+++ b/src/crawlers.ts
@@ -308,21 +308,29 @@ async function chargeStandby(eventName: string, actorRequestId: string): Promise
  * Dispatches to the correct charging path (normal single-run vs. multi-tenant standby)
  * based on isActorStandby().
  */
+const CHARGE_TIMEOUT_MILLIS = 5_000;
+
 async function maybeCharge(crawlerType: ContentCrawlerTypes, actorRequestId?: string) {
     if (getMiniActor().name !== 'url-to-markdown') {
         return;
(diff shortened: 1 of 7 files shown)


src/crawlers.ts
```diff
@@ -292,7 +296,7 @@ async function chargeStandby(eventName: string, userAuthorization: string): Prom
             Authorization: `Bearer ${token}`,
             'Idempotency-Key': randomId(),
         },
-        body: JSON.stringify({ eventName, count: 1, userAuthorization }),
+        body: JSON.stringify({ eventName, count: 1, requestId: actorRequestId }),
     });
```
nikitachapovskii-dev, line 300: As discussed in slack we would better add timeout here

nikitachapovskii-dev (commented):
Thanks for explanation! Please fix lint alert `Error:   10:45  error  'sendResponseIfFinished' is defined but never used  @typescript-eslint/no-unused-vars`
and the point below:

nikitachapovskii-dev (approved):
No objections!
I think the fix for media can be done in a separate pr if we want to fix it 👍

PR #143, opened 2026-08-26
fix: warn when browser-playwright is used with low memory

### LINKED ISSUE #142: Playwright scraping tool times out far more often at 1024 MB memory
## Background

Analyzed actor usage/run data (Redash/Snowflake, `APIFY_ANALYTICS`) for RAG Web Browser covering 2026-05-13 to 2026-08-25 (~3.85M runs, 101,834 distinct users). Overall reliability is excellent (99.44% success), but one config combination stands out.

For runs using `scrapingTool=browser-playwright`, timeout rate by allocated run memory:

| Memory | Runs | Timeout % | Success % |
|---|---|---|---|
| 1024 MB | 8,057 | **35.6%** | 63.8% |
| 2048 MB | 3,294 | 1.3% | 96.9% |
| 4096 MB | 5,640 | 1.5% | 97.0% |
| 8192 MB | 90,970 | 0.9% | 98.5% |

At 1024 MB, over a third of Playwright runs time out. At 2048 MB and above, the timeout rate drops to roughly 1%.

## Root cause

This isn't memory exhaustion — `stats_memory_max_mbytes` stays well under the 1024 MB cap even for timed-out runs. It's CPU starvation: Apify ties the CPU share to the memory tier, so a Playwright/Firefox browser at the 1024 MB tier gets too little CPU to render a page and extract content before `requestTimeoutSecs` (and the overall run timeout) is hit. The `raw-http` (Cheerio) path doesn't have this problem — it stays near-0% timeout regardless of memory, since it never launches a browser.

Concretely, for the `browser-playwright` + `outputFormats=["text"]` + `maxResults=1` cohort (9,303 runs, 246 users), 24.6% of all runs timed out — driven almost entirely by users on the 1024 MB tier.

## Proposal

Since Playwright is memory/CPU-bound in a way raw-http isn't, users who pick `browser-playwright` at the default/low 1024 MB memory tier are silently signing up for a much higher failure rate. Warn them at run start (via `log.warning`) when `scrapingTool=browser-playwright` and allocated memory is at or below 1024 MB, suggesting at least 2048 MB and ideally 4096 MB.

A warning is sufficient for now — no automatic memory bump.

nicklamonov:
## Summary
- Fixes #142
- Playwright runs at the 1024 MB memory tier time out ~35.6% of the time vs ~1% at 2048 MB+ (see issue for the full usage-data breakdown). This isn't memory exhaustion — it's CPU starvation, since Apify ties CPU share to the memory tier and a headless browser can't render/extract a page fast enough on 1024 MB's CPU allocation.
- Adds a `log.warning` at run start when `scrapingTool=browser-playwright` and the run's allocated memory (`Actor.getEnv().memoryMbytes`) is ≤ 1024 MB, suggesting at least 2048 MB, ideally 4096 MB. No automatic memory change — warning only, per the agreed scope.

## Test plan
- [x] `pnpm exec tsc --noEmit` — clean
- [x] `pnpm run lint` — clean
- [x] `pnpm exec vitest run tests/input.test.ts` — 5/5 new tests pass (warns at 1024/512 MB + playwright, silent at 2048 MB+, silent for raw-http, silent when memory unknown)
- [x] `pnpm exec vitest run` — no new failures; the 5 pre-existing `standby.test.ts`/`playwright-crawler.content.test.ts` timeouts reproduce identically on `master` without this change (sandboxed environment lacks a real browser/network), unrelated to this fix

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/input.ts
```diff
@@ -328,6 +329,28 @@ function validateAndFillInput(input: Partial<Input>): Input {
...
+ * Playwright needs meaningfully more memory (and, since Apify ties CPU share to the memory tier, more CPU)
+ * than the raw-HTTP path to render a page within the request timeout. Usage data shows runs on the
+ * `browser-playwright` tool time out ~35% of the time at the 1024 MB tier, vs. ~1% at 2048 MB and above,
+ * so we warn users into bumping memory rather than letting them discover the timeouts themselves.
+ */
+const LOW_MEMORY_FOR_PLAYWRIGHT_MBYTES = 1024;
+const RECOMMENDED_PLAYWRIGHT_MEMORY_MBYTES = 4096;
+const MIN_RECOMMENDED_PLAYWRIGHT_MEMORY_MBYTES = 2048;
+
+export function warnIfLowMemoryForPlaywright(scrapingTool: ScrapingTool, memoryMbytes: number | undefined) {
```
ruocco-l, line 342: I think this is a good candidate to go in the tip object in the kvs

tests/input.test.ts
```diff
@@ -0,0 +1,51 @@
...
+    });
+
+    it('does not warn when using browser-playwright with enough memory', () => {
+        const warning = vi.spyOn(log, 'warning').mockImplementation(() => undefined);
+
+        warnIfLowMemoryForPlaywright('browser-playwright', 2048);
+
+        expect(warning).not.toHaveBeenCalled();
+    });
+
```
ruocco-l, line 35: This could be just one test

ruocco-l (approved):
Just a small suggestion, I'm good with it (but also, we should try to understand why users are lowering the memory, maybe we are communicating something in a wrong way)

tests/input.test.ts
nicklamonov: Consolidated into a single test in b346778.

PR #144, opened 2026-08-27
fix: enqueue google relative urls

ruocco-l:
Fix Google search sometimes returning organic results as relative click-tracking redirects (/goto?url=..., /url?q=...) instead of plain absolute links, which failed to enqueue as crawl requests (see [this run](https://console.apify.com/view/runs/pmNJfI4Pd4Buk2b7l)).

These known Google redirect paths are now resolved against the page they were loaded from before being enqueued — crawling them lets Google's own redirect resolve to the real destination. Other relative/malformed hrefs are now explicitly rejected instead of silently passing validation.

Also prefers the loaded url to the request url for the output


nikitachapovskii-dev (approved):
No regressions found from my side. Maybe we later should cover google-extractors-urls.ts with tests. LGTM

PR #147, opened 2026-09-01
feat: recommend specialized Actors via the TIP key

### LINKED ISSUE #141: Actor tips via reserved TIP key in the KVS
## Summary

Let an Actor surface short, human-readable advisory messages — "tips" — about its own run, using a reserved key-value store key (`TIP`) with the same lightweight convention as `OUTPUT`. No new platform APIs or schema changes required; any consumer that already knows to read `OUTPUT` can equally read `TIP`.

## Motivation

Actors sometimes know something useful about a run that isn't itself a result:

- "You pointed Start URLs at `instagram.com` — there's a specialized Actor built for scraping that site."
- "You set `maxItems` to `1` but `maxConcurrency` to `50` — these don't interact the way you might expect." (this might be more valid for other actors though)

Today there's no place to put this. It doesn't belong in the dataset (a tip isn't a data row and would corrupt exports/schema validation), and there's no existing "advisory" channel on the KVS. This proposal gives Actors a standard, zero-infrastructure way to emit that message, and gives any downstream reader (Console, MCP tool wrapper, apify-client, custom scripts) a predictable place to look for it.

## Proposed solution

- Reserve a KVS key named `TIP` (naming TBD — see Open Questions), written via the SDK the same way `OUTPUT` is today:

```js
await Actor.setValue('TIP', {
    message: 'Start URL matches a site with a dedicated Actor: apify/website-content-crawler.',
    level: 'info', // e.g. info | warning
});
```

- No schema is enforced platform-side, matching the `OUTPUT` precedent — the shape above is a suggested convention, not a contract.
- Readers fetch it exactly like any other KVS record: `GET /v2/key-value-stores/:id/records/TIP`, or via `client.keyValueStore(id).getRecord('TIP')`.
- Consumers decide what to do with it. In particular, an MCP-tool wrapper around Actor runs can prepend the tip as a `content` text block ahead of the dataset-items result, since `content` is the part MCP clients reliably forward into model context.

## What this does and doesn't cover

- ✅ Actor-authored tips (contradictory inputs, ecosystem suggestions, anything the Actor's own code can detect).
- ❌ Platform-generated tips (e.g. "this run OOM'd, try more memory") — the Actor process can't reliably self-report if it was killed, so that case needs a separate platform-side writer feeding the same key, or a different delivery route entirely. Out of scope for this proposal.

## Example end-to-end flow

1. Actor code detects an issue during a run and calls `Actor.setValue('TIP', {...})`.
2. Run finishes; the KVS record persists as normal.
3. A consumer (e.g. the MCP server wrapping this Actor as a tool) fetches the `TIP` record after the run completes and surfaces it however fits — inline text before results, a Console banner, etc.

## Open questions

- **Naming.** `TIP` vs. `TIPS` (singular vs. list-shaped) vs. something else — also needs to avoid colliding with the existing "Actor Issues" feature conceptually.
   - Recommendation: `TIP`
- **Single value vs. list.** Should this support multiple tips per run, or just the latest overwrite (like `statusMessage`)?
   - Recommendation: Single value
- **Suggested schema.** Do we publish a recommended shape (`message`, `level`, maybe `code`) so consumers can rely on some structure, or leave it fully freeform like `OUTPUT`?
- **Discoverability.** How do consumers know an Actor supports tips at all — documentation convention only, or something declared in `actor.json`?
   - Will be decided after implementation as next step.

## Non-goals

- No changes to `apify-core`, the Run entity, or dataset-item response shapes.
- No guarantee of visibility outside consumers that explicitly choose to read the `TIP` key.

## List of actors to recommend:
Facebook 
- for posts: https://apify.com/apify/facebook-posts-scraper
- for groups: https://apify.com/apify/facebook-groups-scraper

[zillow.com](http://zillow.com/) 
- https://apify.com/maxcopell/zillow-detail-scraper

[Instagram.com](http://instagram.com/) 
- https://apify.com/apify/instagram-scraper

[google.com](http://google.com/)
- For Google maps requests: https://apify.com/compass/crawler-google-places

[booking.com](http://booking.com/) 
- https://apify.com/voyager/booking-scraper

[tripadvisor.com](http://tripadvisor.com/)
- https://apify.com/maxcopell/tripadvisor

Youtube 
- https://apify.com/streamers/youtube-scraper

Tiktok 
- https://apify.com/clockworks/tiktok-scraper

[amazon.com](http://amazon.com/) 
- https://apify.com/junglee/amazon-crawler

Example phrase when the criteria is met: "For scraping facebook.com, we recommend using [Facebook Posts Scraper](https://console.apify.com/actors/KoJrdxJCTtpon81KY)"

nikitachapovskii-dev:
In normal mode the Actor now writes an advisory record under the reserved KVS key `TIP` when the query or URL targets a site that has a dedicated Actor:

{ "message": "For scraping tiktok.com, we recommend using [TikTok Scraper](https://apify.com/clockworks/tiktok-scraper)", "level": "info" }

Covers exactly the sites listed in the issue. Matches both plain URLs and domains mentioned in a search query (`site:instagram.com nike`).

Note: the record is stored **after** the crawl. Crawlee purges the default storages while the crawlers start up, so a write placed before that is silently lost.

Closes #141.


src/tips.ts
```diff
@@ -0,0 +1,151 @@
...
+    },
+    {
+        site: 'amazon.com',
+        actorTitle: 'Amazon Crawler',
+        actorUrl: 'https://apify.com/junglee/amazon-crawler',
+        domains: ['amazon.com'],
+    },
+];
+
+const SURROUNDING_QUOTES = /^["']+|["']+$/g;
```
RobertCrupa: Do we not want to strip punctuation too? for example:

"go to zillow.com,"

RobertCrupa (approved):
Are we okay with only .com TLDs?

www.google.co.uk will fail for example

RobertCrupa (approved):
Nice one!

src/tips.ts
nikitachapovskii-dev: Good points, thanks a lot 👍

Stripped surrounding punctuation before parsing

Also added TLDs (used in crawlee) so a rule can now say domains: ['amazon.*'], which means the site under any public suffix. So ***co.uk work now. 
I only applied it to google, amazon and tripadvisor cause claude says other seven have a single global domain 🤷‍♂️

PR #148, opened 2026-09-03
feat: pay-per-event pricing

### LINKED ISSUE #146: Migrate pricing model from PPU to Pay-Per-Event (PPE)
## Summary

Move RAG Web Browser from its current compute-unit ("PLATFORM"/PPU) pricing to pay-per-event (PPE). Pricing below is **approved** — this issue tracks the implementation.

## Background

RAG Web Browser has never left its original PPU pricing (`actor_paid_pricing` has zero rows for it). Since Q1 2026 it has become one of Apify's highest-*volume* actors almost entirely because of Apify MCP: monthly distinct users went from ~1,700 (Jan 2026) to ~37,000–41,000 (Mar–Jul 2026), with MCP now ~72% of runs and ~90%+ of the user base. That growth is heavily skewed toward tiny, cheap, often one-off runs — revenue grew only ~3x while runs grew ~12x.

Two additional reasons beyond revenue:
- **Multi-tenant Standby mode requires moving to PPE** — Apify's Standby infrastructure is heading toward a multi-tenant model, and that model needs per-event charging to bill individual callers correctly rather than one shared run. This actor's real-time API traffic can't move to multi-tenant Standby without PPE in place first.
- **Agentic payment protocols (x402/Skyfire) require PPE pricing** to be eligible at all — a PPU actor is structurally excluded regardless of how well it's priced. RAG Web Browser's heavy agentic/MCP traffic is exactly the population this matters for.

Current blended margin is 84.95% (via `DATAMART_KEBOOLA.actor_run_platform_usage_cost_revenue_detail`). Full research/modeling detail (usage patterns, cost breakdown, competitor pricing, per-tier margin projections) lives in the internal report: https://app.notion.com/p/apify/RAG-Web-Browser-move-to-PPE-pricing-3b3f39950a2280a29213f5a89f206feb

## Pricing grid 
Three flat events, applied identically in Standby and Normal mode **except** Actor Start, which is Normal-mode only:

| Event | Bronze price (anchor) |
|---|---|
| Fetch (page extraction, both modes) | $1.50 / 1,000 |
| Actor Start (Normal mode only, once per run) | $0.70 / 1,000 |
| Search (both modes, once per search run) | $1.50 / 1,000 |

**Tier schedule** — Free priced *above* Bronze (standard platform pattern, incentivizes signing up for a paid plan):

| Tier | Multiplier | Fetch | Actor Start | Search |
|---|---|---|---|---|
| Free | 1.20x | $1.800 | $0.840 | $1.800 |
| Bronze (anchor) | 1.00x | $1.500 | $0.700 | $1.500 |
| Silver | 0.85x | $1.275 | $0.595 | $1.275 |
| Gold | 0.70x | $1.050 | $0.490 | $1.050 |
| Platinum | 0.55x | $0.825 | $0.385 | $0.825 |
| Diamond | 0.40x | $0.600 | $0.280 | $0.600 |

Illustrative per-run prices (per 1,000 runs; search example = 1 search + 3 fetches, the observed average):

| Tier | Standby fetch | Normal fetch | Standby search | Normal search |
|---|---|---|---|---|
| Free | $1.80 | $2.64 | $7.20 | $8.04 |
| Bronze | $1.50 | $2.20 | $6.00 | $6.70 |
| Silver | $1.28 | $1.87 | $5.10 | $5.70 |
| Gold | $1.05 | $1.54 | $4.20 | $4.69 |
| Platinum | $0.83 | $1.21 | $3.30 | $3.69 |
| Diamond | $0.60 | $0.88 | $2.40 | $2.68 |

On July 2026 volumes this raises revenue ~+12–16% above the current PPU baseline ($11,806.23) at ~85% margin (vs. current 83.30%).

## Related input/config decisions (already made, in scope for this migration)

- **`serpProxyGroup`**: hide this input from the UI. Full deprecation can follow later — do not charge for it separately.
- **`proxyConfiguration`**: keep as-is for now. No separate residential-proxy surcharge event at launch — watch usage/cost and introduce one later if needed.
- **`scraping_tool` (raw-http vs. browser-playwright)**: no separate browser-rendering surcharge at launch, for the same reason — revisit later if cost divergence warrants it.

## Rollout mechanics

- 14-day notice period applies platform-wide to any pricing model or price change — not configurable.
- Migration path: switch pricing model → 2-week buffer → ship a build with conditional `actor.charge()` calls that work under both old and new pricing → delete builds that predate the charge calls (they'd otherwise run free under PPE) → buffer ends → drop the conditional logic. 
- Per-user price overrides (Admin Settings → Add pricing override, ~1-year validity) are available if any high-touch account needs protecting from this change — no engineering required.

## Scope of this issue

- [ ] Implement `actor.charge()` calls for Fetch, Actor Start, Search events per the grid above
- [ ] Ensure Standby PPE support too
- [ ] Hide `serpProxyGroup` input
- [ ] Set up tiered pricing (Free/Bronze/Silver/Gold/Platinum/Diamond) per the multiplier table
- [ ] Coordinate 14-day pricing-model-change notice
- [ ] Ship dual-compatible build, then buffer, then drop legacy builds per rollout mechanics above (**Is it possible?**)

nikitachapovskii-dev:
Moves RAG Web Browser to pay-per-event pricing with three events: actor-start (Normal mode only, once per run), search (once per query, both modes) and fetch (once per extracted page, both modes).

Charging moves into src/charging.ts, driven by per-mini-actor chargeEvents. In Standby the charge is attributed to the caller through x-actor-request-id and is always awaited before the response is sent, since the request ID is only valid while its request is in flight. 

URL to Markdown's charging is deliberately left as it behaves today: same events, same moment of charging, same skip when a Standby request carries no request ID. The one place the two Actors differ is that no-request-ID branch: for single-tenant Standby the run belongs to the caller, so charging its owner is still correct, while for multi-tenant it would bill us. standbyTenancy in src/mini-actors.ts keeps them apart.

Also hides the serpProxyGroup input from the UI. its enum had to go, because a string field carrying an enum may only use editor: "select" and rewrites the README pricing section, which still claimed the Actor was free of charge.

Verified with local runs under ACTOR_TEST_PAY_PER_EVENT: a normal run charges actor-start + fetch, a search run adds search, and a run on the current compute-unit pricing charges nothing at all.

Needs Console setup. The platform's automatic apify-actor-start event must be set to $0, otherwise it duplicates Actor Start.

Closes #146


src/charging.ts
```diff
@@ -0,0 +1,123 @@
...
+     * being charged for rather than the attempt - `Request.uniqueKey` does exactly that.
+     */
+    idempotencyKey: string;
+}
+
+/**
+ * Whether the current run is billed per event. Charging is a no-op otherwise, so one build works both
+ * before and after an Actor is switched over to pay-per-event pricing.
+ */
+function isPayPerEvent(): boolean {
```
ruocco-l, line 25: Feels like a redundant check? How is this useful?

src/charging.ts
```diff
@@ -0,0 +1,123 @@
...
+    try {
+        return Actor.getChargingManager().getPricingInfo().isPayPerEvent;
+    } catch {
+        // `getPricingInfo` throws until `Actor.init()` has initialized the charging manager.
+        return false;
+    }
+}
+
+async function chargeRunOwner(eventName: string): Promise<void> {
```
ruocco-l, line 35: This could be inlined

src/charging.ts
```diff
@@ -0,0 +1,123 @@
...
+}
+
+/**
+ * Charges the caller of a live STANDBY request by POSTing to the platform charge endpoint with the ID
+ * of that request, so that the caller is billed rather than the owner of the (possibly shared) run.
+ *
+ * The request ID is only valid while its HTTP request is in flight, so this must be awaited before
+ * the response to that request is sent.
+ */
+async function chargeStandbyCaller(eventName: string, actorRequestId: string, idempotencyKey: string): Promise<void> {
```
ruocco-l, line 50: This is ok, but it's not clear from the function itself that this is meant for Multi-tenant (server) standby, let's clarify it

src/charging.ts
ruocco-l: Or we can keep it and race this against a time out so that the timeout lives in the endpoint call like it is for `chargeStandbyCaller`

src/mini-actors.ts
```diff
@@ -23,6 +51,14 @@ const MINI_ACTORS: Record<string, MiniActor> = {
         mcpServerName: 'mcp-server-rag-web-browser',
         route: Routes.SEARCH,
         helpRoute: '/search?query=hello+world',
+        standbyTenancy: 'SINGLE_TENANT',
+        chargeEvents: {
+            search: 'search',
+            fetch: {
+                [ContentCrawlerTypes.CHEERIO]: 'fetch',
+                [ContentCrawlerTypes.PLAYWRIGHT]: 'fetch',
```
ruocco-l, line 60: Why are we not separating the event based on the crawler like we do on URL2M?

ruocco-l (approved):
Just a bunch of NITs and a question on the events. I am still not entirely ok with the tests, but it's something that we should not try to correct here.

src/charging.ts
nikitachapovskii-dev: No, it's for the migration window, when we wait for 14-day notice. But it's useless where I put it indeed, I would better move it into chargeStandbyCaller.
Without the check the POST just gets 400 cannot-charge-non-pay-per-event-actor and we log a charging error on every Standby request until the pricing switches.

src/charging.ts
nikitachapovskii-dev: Yeah, went with your second option 👍

src/charging.ts
nikitachapovskii-dev: Updated the comment ✅

src/mini-actors.ts
nikitachapovskii-dev: idk why but it's one of issue instructions 😄 we can discuss it tomorrow

Screenshot 2026-09-07 at 16 22 54

src/charging.ts
```diff
@@ -0,0 +1,125 @@
...
+        return Actor.getChargingManager().getPricingInfo().isPayPerEvent;
+    } catch {
+        // `getPricingInfo` throws until `Actor.init()` has initialized the charging manager.
+        return false;
+    }
+}
+
+/** Charges the owner of the run. Unlike the endpoint call below, the SDK call has no deadline. */
+async function chargeRunOwner(eventName: string): Promise<void> {
+    await withTimeout(Actor.charge({ eventName }), CHARGE_TIMEOUT_MILLIS);
```
ruocco-l, line 36: Instead of having asymmetry and flags all over because this method does not support the idempotency key, why don't we just do both HTTP requests for charging and switch to the method once everything is supported for multi-tenant?

nikitachapovskii-dev: > just do both HTTP requests for charging

That would cost us more than it saves because we also have to think about:
1) Actor.charge stops charging once the run hits maxTotalChargeUsd so over the endpoint we'd keep billing past the user's limit.
2) It also retries with backoff 🫠

